PR #1, opened 2024-10-01
Annotation model and basic FileStorage system arch

Synicix:
Work in progress:

I implemented the basic architecture for annotation and the store class. Still need to write the test cases then start working on pods since we change the spec on that a bit.


eywalker (commented):
Coming in late but looks like a great starting point! Let's also start placing skeletons of methods that must be implemented so that we can define the signature fully

PR #6, opened 2024-10-02
First implementation round of pod and annotation with store

### LINKED ISSUE #10: Add `Pod` model
```rust
struct Annotation {
    api_ver: String // Adopt the version lib
    name: String,
    version: String,  // better option could be used later for inequalities
    description: String,
}

struct Pod {
    annotation: Annotation,
    command: String,
    gpu: Option<GPUSpecRequirement>,
    image_digest: String, // resolved docker image+digest
    input_stream_map: BTreeMap<String, PathBuf>,
    // might default to /output but needs
    output_dir: PathBuf,
    // here PathBuf will be relative to output_dir
    output_stream_map: BTreeMap<String, PathBuf>,
    recommended_cpus: f32,  // e.g. 0.25
    recommended_memory: u64, // e.g. 2GiB -> 2147483648 bytes
    sha256: String,  // pod hash, might be a better type
    source: String, // Git Commmit permalink
}

impl Pod {
    fn new(...) -> Self {...}
    
// Use the new derivative default lib
}



impl OrcaHash for Pod {
    fn to_yaml(&self) -> String
   pub fn compute_hash(&self) -> String
  fn from_yaml(yaml: &str, annotation: &str) -> Self
} 
```

### LINKED ISSUE #16: OrcaStore and LocalFileStore
```rust
trait OrcaStore {
    fn save_pod(&self, pod: Pod) -> Result<(), String>;
    fn load_pod(&self, sha256: String) -> Option<Pod>;
    fn delete_pod(&self, sha256: String) -> Result<(), String>;
    // key=column, value Vec of values e.g. name=[a,b], version=[0.1.0,1.0.0],sha256=[abcd,dcba] 
    fn list_pod(&self) -> BtreeMap;  
}

struct LocalFileStore {
    pub location: PathBuf,
}

impl OrcaStore for LocalFileStore {...};
```

### LINKED ISSUE #17: OrcaHash trait
```rust
trait OrcaHash {
  fn to_yaml(&self) -> String,
  fn from_yaml(yaml: &str, annotation: &str) -> Self,
  fn compute_hash(&self) -> String {//provide default implementation}
}
```

Synicix:
Initial implementation of core types in Orcapod with file storage implementation


Synicix:
Here is the first iteration of the code base for Pod and Annotation, along with store.

There are still a few things I plan on adding
- Better testing function that handles cleanup correctly @guzman-raphael You are an expert on this so any advice would be helpful
- Missing docker image checking and pulling (Need to copy my code from the old base to this new one)

As for the yaml conversion thing I post on slack, @guzman-raphael @eywalker , for serialization it works amazing, however trying to do something similar to Rapahel's solution for deserialization, converting the Values (from Serede_yaml) back to our struct is a pain in the butt. So as a stop gap, I just went back to my yaml struct for deserialization only and wrote it in a way where the data only get created once and gets moved. If there is a better solution, I am all ears.

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+        Ok(pod)
+    }
+}
+
+/// Helper Functions
+///
+
+fn create_file_and_dir_if_not_exist(path: &Path, content_to_write: &str) -> Result<(), String> {
+    match Path::new(&path).exists() {
+        true => Err(format!(
```
eywalker: Simple if-else would be cleaner for bool check

src/model.rs
```diff
@@ -0,0 +1,158 @@
...
+        pod
+    }
+
+    pub fn compute_hash(&self) -> String {
+        compute_sha_256_hash(&self.to_yaml())
+    }
+
+    pub fn verify(&self) -> Result<(), String> {
+        let compute_hash = self.compute_hash();
+        match compute_hash != self.pod_hash {
```
eywalker: let's use if/else for bool pattern

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+};
+
+static POD_FOLDER_NAME: &str = "Pod";
+
+pub struct FileStore {
+    storage_folder_path: PathBuf,
+}
+
+impl FileStore {
+    pub fn new(storage_folder_path: PathBuf) -> Self {
```
eywalker: let's make it more generic and take in anything that can be turned into PathBuf. So more appropriate will be 
```rust
pub fn new(storage_folder_path: impl Into<PathBuf>) -> Self
```

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+        Self {
+            storage_folder_path: storage_folder_path,
```
eywalker: With the change suggested above, this would be
```rust
storage_folder_path: storage_folder_path.into()
```

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    fn construct_annotation_path(&self, hash: &str) -> PathBuf {
+        let mut path_buf = self.storage_folder_path.to_path_buf();
```
eywalker: I'd think `clone()` would be more appropriate than `to_path_buf`

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+        path_buf.push(format!("{}.yaml", hash));
+
+        path_buf
+    }
+}
+
+impl Store for FileStore {
+    fn store_annotation(
+        &self,
+        annotation: &crate::model::Annotation,
```
eywalker: please add `use` of Annotation for consistency and avoid fully qualified name

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+        hash: &str, // Of owner
+    ) -> Result<(), String> {
+        let yaml_str = annotation.to_yaml();
+
+        let path = self.construct_annotation_path(hash);
+
+        match create_file_and_dir_if_not_exist(&path, &yaml_str) {
```
eywalker: since we don't handle it explicitly we could just do `create_file_and_dir_if_not_exist(...)?` right?

src/filestore.rs
```diff
@@ -0,0 +1,224 @@
...
+        match create_file_and_dir_if_not_exist(&path, &pod.to_yaml()) {
+            Ok(_) => (),
+            Err(e) => return Err(e.to_string()),
+        };
+
+        // Missing docker image save
+        // TODO
+
+        // Save the Annotation
+        match create_file_and_dir_if_not_exist(
```
eywalker: use `store_annotation` method implemented above

eywalker (approved):
Looks good -- a few changes being requested in interface/styling. Also please start adding docs.

src/filestore.rs
Synicix: I actually still need to look up the correct way to pass around path. It seems that in std they use something like this:
```
pub fn some_public_func<P: AsRef<Path>>(path: P) -> ... {
    some_public_func_imp(path.as_ref())
}

fn _some_public_func(path: &Path) -> ... { ... }```
```

src/filestore.rs
Synicix: Ah that was stupid vs code auto doing that

src/filestore.rs
Synicix: Ah yes, I didn't change it after redesigning the signature for the above function to return the same result pattern

Synicix:
@guzman-raphael @eywalker 

With Rapahel feedback on Monday evening, I have change worked on majority of the changes needed to make the code more elegant.

The only thing I am missing from the feedback is the changing of the storage path for Annotation and the pod save function since it is not possible to uniquely identify a pod with multiple annotation attached to it with just the hash.

We had a debate on this and would like your opinion @eywalker on the function signature.
Current way uses pod_hash + annotation_ver to load the pod,
@guzman-raphael suggested having the user access via pod_name + version

Not sure which one you think would be more intuitive for the user? I am starting to lean towards @guzman-raphael suggestion cause most people use docker image by name and not by hash.



Also on the file permission, 644 doesn't play nice with my system but 744 does.

Cargo.toml
```diff
@@ -5,3 +5,8 @@ edition = "2021"
 license = "MIT license"
 
 [dependencies]
+colored = "2.1.0"
+fmt = "0.1.0"
```
guzman-raphael: This doesn't appear to be necessary.

src/utils.rs
```diff
@@ -0,0 +1,8 @@
+pub fn get_struct_name<T>() -> String {
+    std::any::type_name::<T>()
+        .split("::")
+        .collect::<Vec<&str>>()
+        .get(2)
+        .unwrap()
```
guzman-raphael: We should avoid use of `unwrap()` since it is unsafe. Rather we should propagate the error and treat this as a `Result` if it can err.

src/orca.rs
```diff
@@ -0,0 +1,43 @@
+use std::error::Error;
```
guzman-raphael: I remember us discussing this and you mentioned that it is meant to be a reference for the user interface i.e. eventually web service + CLI. Since it is not in scope for the current milestones, I'd suggest we separate this into its own issue/PR.

tests/model.rs
```diff
@@ -0,0 +1,92 @@
+use std::{collections::BTreeMap, fs, path::PathBuf};
+
+use orcapod::{
+    model::{KeyInfo, Pod, PodNewConfig},
+    orca::{create_pod, load_pod, StorageBackend},
```
guzman-raphael: Corresponding to my previous comment about separating `orca` mod into a separate issue/PR, we'd need to restructure this integration test to use other mods to verify behavior.

src/filestore.rs
```diff
@@ -0,0 +1,182 @@
+use std::{
```
guzman-raphael: Rather than defining `store` and `filestore` in a flat design. I'd suggest we either keep it simple within a single `store.rs` or make `store` a directory module with `filestore` as a subdirectory. Seems like this is the best practice for package layout from docs I found.

Refs
- https://doc.rust-lang.org/cargo/guide/project-layout.html
- https://dev.to/ghost/rust-project-structure-example-step-by-step-3ee

src/store.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    fn store_annotation(
+        &self,
+        annotation: &Annotation,
+        owner_hash: &str,
+    ) -> Result<(), Box<dyn Error>>;
+
+    fn load_pod(&self, hash: &str, version: &str) -> Result<Pod, Box<dyn Error>>;
+
+    fn store_pod(&self, pod: &Pod) -> Result<(), Box<dyn Error>>;
```
guzman-raphael: This differs from the corresponding issue spec (#16). Mainly
- `Store` -> `OrcaStore`
- `store_pod` -> `save_pod`
- `list_pod` missing
- `delete_pod` missing

src/model.rs
```diff
@@ -0,0 +1,118 @@
...
+            return Err(Box::new(SerializeError {
+                item_debug_string: format!("{:?}", item),
+                error,
+            }))
+        }
+    };
+
+    yaml_str.insert_str(0, &format!("class: {}\n", get_struct_name::<T>()));
+
+    Ok(yaml_str)
```
guzman-raphael: This is missing a sort of the fields since it assumes that struct defines them in lex order (no guarantee in the future). Think this is the best place to apply the sort since it is specific to generating consistently hashed yaml. 

Have a look at the implementation [here](https://github.com/walkerlab/orcapod/pull/5/files#diff-6bd13171e8498a8adc17d13cef1f1b59f5fbd24ae4c72aeb4ce5f9b2b4703220R11-R22) since I found a way to achieve this and keep the model structs clean by *skipping* the `annotation` and `hash` fields too.

src/filestore.rs
```diff
@@ -0,0 +1,182 @@
...
+            description: yaml_struct.description,
+            version: yaml_struct.version,
+        })
+    }
+
+    fn store_pod(&self, pod: &Pod) -> Result<(), Box<dyn Error>> {
+        let path = self.construct_folder_path(&get_struct_name::<Pod>(), &pod.pod_hash);
+
+        // Try to save it
+        create_file_and_dir_if_not_exist(&path, &to_yaml(&pod)?)?;
```
guzman-raphael: Seems like this will return an error and halt if the pod spec already exists. I agree that this makes sense on the annotation file but not on the spec file.

Consider the case where a user has already created a pod and simply updates the annotation. The pod spec (and hash) will be reused and only a new annotation file needs to be written. Therefore, it shouldn't return an error if the pod spec already exists (maybe just skip and log it) and continue to write the annotation file. 

Have  a look at how I approached this [here](https://github.com/walkerlab/orcapod/pull/5/files#diff-41a991785a2eba02345db5fdcbf15b660f27b2841b8f61a084a7db66e69bf3eeR111-R118).

src/filestore.rs
```diff
@@ -0,0 +1,182 @@
...
+    Ok(match fs::read_to_string(&path) {
+        Ok(raw_yaml) => raw_yaml,
+        Err(error) => {
+            return Err(Box::new(IOError {
+                path: path.as_ref().into(),
+                error,
+            }))
+        }
+    })
+}
```
guzman-raphael: We should probably compare notes on this. I found a way to implement `from_yaml` [here](https://github.com/walkerlab/orcapod/pull/5/files#diff-6bd13171e8498a8adc17d13cef1f1b59f5fbd24ae4c72aeb4ce5f9b2b4703220R24-R43).

src/filestore.rs
```diff
@@ -0,0 +1,182 @@
...
+        let path = self.construct_annotation_path(hash, version);
+
+        let file = read_yaml(&path)?;
+
+        #[derive(Deserialize)]
+        struct AnnotationYaml {
+            name: String,
+            description: String,
+            version: String,
+        }
```
guzman-raphael: Unclear why this was needed.

src/filestore.rs
```diff
@@ -0,0 +1,182 @@
...
+            gpu: Option<GPUSpecRequirement>,
+            image_digest: String,
+            // Num of recommneded cpu cores (can be fractional)
+            input_stream_map: BTreeMap<String, KeyInfo>,
+            min_memory: u64,
+            output_dir: PathBuf,
+            output_stream_map: BTreeMap<String, KeyInfo>,
+            recommended_cpus: f32,
+            source_commit: String, // Git Commit
+        }
```
guzman-raphael: We should compare notes. I was able to avoid this by my use of a `from_yaml` generic method (see [here](https://github.com/walkerlab/orcapod/pull/5/files#diff-41a991785a2eba02345db5fdcbf15b660f27b2841b8f61a084a7db66e69bf3eeR171-R175)).

src/filestore.rs
```diff
@@ -0,0 +1,182 @@
...
+    path: impl AsRef<Path>,
+    content_to_write: &str,
+) -> Result<(), Box<dyn Error>> {
+    if path.as_ref().exists() {
+        return Err(Box::new(FileAlreadyExists {
+            path: path.as_ref().into(),
+        }));
+    } else {
+        // Create the all the folders above the file
+        let parent_path = match path.as_ref().parent() {
```
guzman-raphael: Rather than `match`, you can use `ok_or(err)`. It is pretty neat. It will convert an option to a result and for the `None` case will return the `err` you pass in. That is pretty much exactly what you are doing here. I actually use it quite a bit ([here](https://github.com/walkerlab/orcapod/pull/5/files#diff-41a991785a2eba02345db5fdcbf15b660f27b2841b8f61a084a7db66e69bf3eeR88-R90) is an example).

src/model.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub version: String,
+    pub input_stream_map: BTreeMap<String, KeyInfo>,
+    pub output_dir: PathBuf,
+    pub output_stream_map: BTreeMap<String, KeyInfo>, // Will be relative path of output_dir
+    pub image_name: String,
+    pub source_commit: String,         // Git Commit only for reference
+    pub recommended_cpus: Option<f32>, // Num of recommneded cpu cores (can be fractional)
+    pub min_memory: Option<u64>,       // Bytes
+    pub gpu_spec_requirments: Option<GPUSpecRequirement>,
+}
```
guzman-raphael: Not a big fan of this approach since it still feels quite similar to explicitly defining all of these in a `new(...)` but with added confusion since user would need to interact with both and it makes the codebase a bit harder to follow as the models grow. Since the default is still set in `Pod::new` anyway, I'd prefer to just defining them there. Have a look how I approached this [here](https://github.com/walkerlab/orcapod/pull/5/files#diff-6bd13171e8498a8adc17d13cef1f1b59f5fbd24ae4c72aeb4ce5f9b2b4703220R71-R109).

src/model.rs
```diff
@@ -0,0 +1,118 @@
...
+            },
+            image_digest: config.image_name,
+            recommended_cpus,
+            min_memory: min_memory,
+            gpu_spec_requirments: config.gpu_spec_requirments,
+            source_commit: config.source_commit,
+        };
+
+        // Covert to yaml, if fails, just panic since it should always be valid
+        let pod_hash = format!("{:X}", Sha256::digest(to_yaml(&pod).unwrap()));
```
guzman-raphael: I'd suggest wrapping the hash into our own method since it will be called later by all the models. It will make it easier to update later if we ever change the hash algo. Here's an [example](https://github.com/walkerlab/orcapod/pull/5/files#diff-4007187965c1fea4fd252a76f5f1007b2ea59f6d86e2f454edee547c5036be24R15-R17).

guzman-raphael (changes requested):
Thanks for addressing my previous suggestions. :handshake:

Had some more time to go through it today and provide new feedback.

src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+            f,
+            "{}{}",
+            "Unable to extract folder path".bright_red(),
+            &self.path.to_string_lossy().bright_cyan(),
+        )
+    }
+}
+
+#[derive(Debug)]
+pub struct FileAlreadyExists {
```
eywalker: Does it make sense for us to implement Error struct specific to such detailed case? Wouldn't model like that of IOError suffice?

src/filestore.rs
eywalker: Agreed with structuring it as a hierarchical

Cargo.toml
Synicix: Remove fmt

src/utils.rs
Synicix: TODO

src/orca.rs
Synicix: agree to remove it for later

tests/model.rs
Synicix: TODO, update the test

src/filestore.rs
Synicix: TODO

Synicix: leave for @guzman-raphael

src/filestore.rs
Synicix: TODO

src/filestore.rs
Synicix: TODO

src/model.rs
Synicix: merge into new TODO

eywalker (changes requested):

guzman-raphael (approved):

src/model.rs
```diff
@@ -0,0 +1,122 @@
...
+        output_dir: PathBuf,
+        output_stream_map: BTreeMap<String, StreamInfo>,
+        recommended_cpus: f32,
+        recommended_memory: u64,
+        required_gpu: Option<GPURequirement>,
+    ) -> Result<Self, Box<dyn Error>> {
+        let pod_no_hash = Self {
+            annotation,
+            hash: String::new(),
+            source_commit: source_commit,
```
eywalker: To be consistent with other lines, we can leave this just `source_commit` right?

src/model.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub model: GPUModel,
+    pub recommended_memory: u64,
+    pub count: u16,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum GPUModel {
+    NVIDIA(String),
+    AMD(String),
+}
```
eywalker: Could you put a short comment on what goes in String? Just easy to forget what's expected here

src/store/filestore.rs
```diff
@@ -0,0 +1,209 @@
+use crate::{
+    error::{FileExists, FileHasNoParent, NoAnnotationFound, NoRegexMatch},
+    model::{from_yaml, to_yaml, Pod},
+    store::Store,
```
eywalker, line 4: For `Store` trait I'd fine relative import makes more sense rather than absolute path

src/store/filestore.rs
```diff
@@ -0,0 +1,209 @@
...
+use std::{collections::BTreeMap, error::Error, fs, iter::Map, path::PathBuf};
+
+#[derive(Debug)]
+pub struct LocalFileStore {
+    pub directory: PathBuf,
+}
+
+impl Store for LocalFileStore {
+    fn save_pod(&self, pod: &Pod) -> Result<(), Box<dyn Error>> {
+        let class = get_struct_name::<Pod>()?;
```
eywalker: What's supposed to be returned here? Wouldn't it always be "Pod"? If so, I find the whole struct name look up an unnecessary step.

src/store/filestore.rs
```diff
@@ -0,0 +1,209 @@
...
+                &pod.annotation.version,
+            ),
+            &serde_yaml::to_string(&pod.annotation)?,
+            true,
+        )?;
+
+        LocalFileStore::save_file(
+            &self.make_spec_path(&class, &pod.hash),
+            &to_yaml::<Pod>(&pod)?,
+            false,
```
eywalker, line 36: Would you mind indicating what's being controlled by true/false? I know it's part of the API for save but otherwise unclear and a quick comment would help a lot

src/store/filestore.rs
```diff
@@ -0,0 +1,209 @@
...
+        )?;
+
+        Ok(())
+    }
+
+    fn load_pod(&self, name: &str, version: &str) -> Result<Pod, Box<dyn Error>> {
+        let class = "pod".to_string();
```
eywalker, line 43: This logic should be used in the save side as well

eywalker (changes requested):

src/store/filestore.rs
Synicix: Save pod is only really meant to return Ok() empty if it is successful. The user should already have a handle of the pod when passing in.

As for get struct_name, we didn't want to hardcode it, hence the get_struct_name function to standardize it.

eywalker (approved):

PR #29, opened 2024-10-15
Add tests

### LINKED ISSUE #19: Scaffold test fixtures
Some tests to consider:
- https://github.com/guzman-raphael/orcapod/tree/2d2c97f2ce38fde089a1be89a66a6d5b8db961fc/tests
- https://github.com/Synicix/orcapod/blob/d71adbe3e349d8e0e295001da8e1c9ca882056d4/tests/model.rs

guzman-raphael:
This PR adds:
- [x] [clippy](https://github.com/rust-lang/rust-clippy): a common Rust linter that helps write idiomatic rust code. There are over 700 lints supported and it is highly configurable. I've configured it to be quite 'annoying' with `allow`s on specific lints so that we can explicitly choose if we don't agree with a lint. The goal is to let the linter handle common mistakes so we can focus more on design and algorithms.
- [x] [llvm-cov](https://github.com/taiki-e/cargo-llvm-cov): a tool to help us run integration tests and measure their code coverage. It seems to be quite powerful. Aside from the code coverage dump to stdout, there is also a `--html` mode which gives a static site which you can explore in depth to understand what is missing. There also seems to be some [support](https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#continuous-integration) to uploading it to [Codecov](https://about.codecov.io/) if we want to add this later.
- [x] `tests` GHA to run the test suite.
- [x] Applies `clippy`'s suggestions. Among them are the specification for fields that were missing in our `Cargo.toml` that seem to be needed for crates.io. Regarding the [categories](https://doc.rust-lang.org/cargo/reference/manifest.html#the-categories-field), they are standard [slugs](https://crates.io/category_slugs) from crates.io. We can pick up to 5 and I left in the comments all that *seem* to be applicable for us. [keywords](https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field) also appears to have a limit of 5 for crates.io.
- Adds the following test fixtures (setup/teardown):
   - [x] a pod using the style-transfer example (`pod_style`)
   - [x] a temporary store (`store_test`)
   - [x] `add_pod_storage` which will attach a test store to a pod.
- [x] A few basic integration tests for `model` and `store`.
- Rename:
   - [x] `get_struct_name` -> `get_type_name`
   - [x] `source_commit` -> `source_commit_url`
   - [x] `min_cpus` -> `recommended_cpus`
   - [x] `min_memory` -> `recommended_memory`
   - [x] `location` -> `directory`
- [x] Use `unwrap()` for impossible case in `get_type_name`
- [x] Fixes some non-default file permissions which are unnecessary.

Fix #19


src/model.rs
```diff
@@ -49,45 +50,45 @@ pub fn from_yaml<T: DeserializeOwned>(
...
-    source_commit: String,
+    source_commit_url: String,
     image: String,
     command: String,
     input_stream_map: BTreeMap<String, StreamInfo>,
     output_dir: PathBuf,
     output_stream_map: BTreeMap<String, StreamInfo>,
-    min_cpus: f32,
-    min_memory: u64,
+    minimum_cpus: f32,
```
Synicix: I believe we need to revisit the discussion point, or perhaps I may have forgot, but I though we agree that min cpus doesn't make much sense? Hence, we left it as recommended cpu?

tests/store.rs
```diff
@@ -0,0 +1,54 @@
...
+        let annotation_file = store.make_annotation_path(
+            "pod",
+            &pod_style.hash,
+            &pod_style.annotation.name,
+            &pod_style.annotation.version,
+        );
+        let spec_file = store.make_spec_path("pod", &pod_style.hash);
+        {
+            let pod = add_pod_storage(pod_style, &store)?;
+            assert!(fs::exists(&spec_file)?);
```
Synicix:
```suggestion
            assert!(spec_file.exists());
```

Using the pathbuf .exist() function might make it look cleaner: https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.exists

src/store/filestore.rs
```diff
@@ -3,10 +3,17 @@ use crate::{
...
+use core::{error::Error, iter::Map};
 use glob::{GlobError, Paths};
 use regex::Regex;
-use std::{collections::BTreeMap, error::Error, fs, iter::Map, path::PathBuf};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
+
+extern crate alloc;
```
Synicix: remove that disable clippy warning

Synicix (changes requested):
After that, can merge

Synicix (approved):

PR #32, opened 2024-10-15
Add store generics for make spec path and make annotation path

Synicix:
@eywalker I know we had a discussion about this last night, but this is related to our idea of making some of the function more generic so we are not generating strings for just label.


Synicix:
Merging it with another pull request probably

PR #33, opened 2024-10-15
Resetting rust analyzer settings to default

Synicix:
@guzman-raphael I found the issue I was mentioning yesterday where the hints were not showing up. I think we should just leave it by default, and you can disable it on your side. There might be other people out there like me who been using it as a default then all of the sudden it is missing, then have to go look for this setting.


guzman-raphael (approved):
@Synicix Makes sense. In hindsight, this does seem set a bit too much for personal preference.

PR #34, opened 2024-10-16
Add genrics store function, modify parse annotation path to search an…

### LINKED ISSUE #26: Make `Store` trait methods generic
`Store` trait has multiple `save` and `load` methods specific to each resource type (e.g. `Pod`, `PodJob`). For a specific implementation, we can simplify the implementation by invoking a generic store/load method that generically works against all of these resource types.

### LINKED ISSUE #27: Make `Annotation` optional
For example, this would make it easier to instantiate a `Pod` and load from store.

Some considerations:
- `load_pod`
  - w/ name/version -> loads annotation + spec into instance
  - w/ hash -> loads spec only into instance
- `save_pod`
  - w/ annotation -> saves both annotation + spec
  - w/o annotation -> saves only spec
- `delete_pod`
  - w/ hash -> delete spec dir + all annotations that refer to it (if annotation dir becomes empty, clean up as well)
  - w/ name/ver -> delete annotation only (if annotation dir becomes empty, clean up as well)
- `list_pod` -> make name and ver `Vec<Option<_>>` to account for `None` on "spec-only" pods

### LINKED ISSUE #36: Change Annotation handling
In the running design, `Annotation` type lives in its own directory, separate from the object it's annotating. This makes moving the target object with annotation difficult. In the new design, `Annotation` will directly live in the resource folder that contains the target resource. For example, an annotation for a `pod` would be found as follows in the filestore:
```
pod
  └──hash
               ├── spec.yaml (Pod resource info)
               └── name-version.yaml  (annotation yaml)

Exact specification of the annotation file name is YTD but multiple annotation files would be allowed.
This issue is related to #27 as a resource may lack any annotation yaml, making annotation optional.

### PR DESCRIPTION
**Pull request by Synicix**
-Add genrics store function
-modify parse annotation path to search annotation
- fix delete pod missing handle case and convert it to use generics
- made get_struct_name panic if fail to extract
- fix regex
- remove regex not found

### CODE REVIEW

**Comment by Synicix** [comment]
#26

**Comment by Synicix** [comment]
@guzman-raphael @eywalker Going to open it up for discussion on either tomorrow or Friday

For the most part it is ready for review and merge, but would like your feedback on how to improve it.

**Comment by Synicix** [comment]
Note #35 will depend on this, since it uses a lot of the generic functions I made

File: src/util.rs [hunk 1]
```diff
@@ -1,15 +1,14 @@
...
 use sha2::{Digest, Sha256};
 use std::any::type_name;
 
-pub fn get_struct_name<T>() -> Result<String, OutOfBounds> {
-    Ok(type_name::<T>()
+pub fn get_struct_name<T>() -> String {
+    type_name::<T>()
         .split("::")
         .collect::<Vec<&str>>()
         .get(2)
```
eywalker: Actually why are we getting position 2?

eywalker (changes requested):
I agree with the strict name lookup not returning a result. However I'd want to reassess its use and logic.

src/util.rs
Synicix: Okay just checked the code, since I didn't really looked at it before after it got merge into the initial pull request.

So, for example, given a pod, type_name::<Pod>() returns "orcapod::model::Pod", so we would need to extract the last section and ignore the rest. Given that, this should have been implemented with last and not get(2). I will make a change quickly

Synicix:
@eywalker @guzman-raphael 
Summary:
- Change annotation design to save in the item folder
- Add cacheing in form of BTreeMap<name-ver, item_hash> with binary encoding and race condition handling for file system io

TODO:
- Need to add glob and regex to construct BTreeMap in case there is a change done outside the system.
- Error Handling (might leave till later)

I have updated the pull request to support the new design of annotation with a slight modification to make the folder structure cleaner.

What we agree to try yesterday was to save annotation in the item folder, but instead of doing that, I now save it under all annotations under its own folder, under the item folder.

i.e
/pod/annotations/<name>-<ver>.yaml
The reason behind this because I didn't want to gunk up the main folder with a bunch of annotations that user may not care about.

As for caching, I implemented a BTreeMap<name-ver, item_hash> along with storing it in a binary and the associated file locking system to prevent race conditions when more than one instance of the program is updating the localfilestore.

.vscode/launch.json
Synicix, line 1: Revert the change

src/store/filestore.rs
```diff
@@ -113,97 +47,256 @@ impl LocalFileStore {
...
-            class,
-            name,
+            get_type_name::<T>(),
             hash,
-            version,
         ))
     }
 
-    fn make_spec_path(&self, class: &str, hash: &str) -> PathBuf {
+    fn make_path<T>(&self, hash: &str, file_name: &str) -> PathBuf {
```
Synicix: Look into join

src/store/filestore.rs
```diff
@@ -113,97 +47,256 @@ impl LocalFileStore {
...
+        let mut path = self.make_dir_path::<T>(hash);
+        path.push(file_name);
+        path
+    }
+
+    fn make_anno_path<T>(&self, hash: &str, file_name: &str) -> PathBuf {
```
Synicix: rename to make_annotation_path

src/store/filestore.rs
```diff
@@ -113,97 +47,256 @@ impl LocalFileStore {
...
+            let mut file = fs::File::open(self.make_name_ver_index_path::<T>())?;
+            let mut temp = Vec::new();
+            file.read_to_end(&mut temp)?;
+        }
+
+        Ok(())
+    }
+
+    /// Generic function for loading spec.yaml into memory
+    fn load_item<T: DeserializeOwned>(
```
Synicix: change to 
load_item<T:asjdfklajsdf>(&self, spec; ResourceSpec)

enum ResourceSpec {
hash(string,
annotation(String, String)
}

src/store/filestore.rs
```diff
@@ -113,97 +47,256 @@ impl LocalFileStore {
...
+        // Get the spec and annotation yaml
+        let spec_yaml = fs::read_to_string(self.make_path::<T>(&hash, "spec.yaml"))?;
+        let annotation_yaml = fs::read_to_string(
+            self.make_anno_path::<T>(&hash, &format!("{}-{}.yaml", name, version,)),
+        )?;
+
+        from_yaml::<T>(&spec_yaml, &hash, Some(&annotation_yaml))
+    }
+
+    fn delete_item<T>(&self, name: &str, version: &str) -> Result<(), Box<dyn Error>> {
```
Synicix: Match ResourceSpec above

Synicix: Explicit delete the item regardless if there are annotation pointing to it.

Synicix: Also add delete annotation function

Synicix: ^ fine with generic

src/store/mod.rs
```diff
@@ -1,10 +1,16 @@
 use crate::model::Pod;
 use std::{collections::BTreeMap, error::Error};
 
+pub struct ItemInfo {
```
Synicix: Be removed

Synicix (commented):
TODO changes

.vscode/launch.json
Synicix: Probably need to be updated

src/store/filestore.rs
Synicix: did join

src/store/filestore.rs
Synicix: reanmed

src/store/filestore.rs
Synicix: updated

src/store/filestore.rs
Synicix: Implemented generic

src/store/mod.rs
Synicix: removed

Synicix:
Ready for review with the new implementation of cache @eywalker @guzman-raphael

.vscode/launch.json
```diff
@@ -33,8 +33,6 @@
                     "test",
                     "--no-run",
                     "--package=orcapod",
-                    "--test",
-                    "model",
```
guzman-raphael: You should modify this on your side as needed but we shouldn't change what is committed here since it is an example. Changes should be only if there is a bug or if there is an improvement to the example.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ sha2 = "0.10.8"
 glob = "0.3.1"
 regex = "1.11.0"
 colored = "2.1.0"
+anyhow = "1.0.90"
```
guzman-raphael: Rather than adding a dependency for this, you could have added an alias with one line in `error.rs`. Then you could import it everywhere.

For example:

```rust
pub type Result<T> = std::result::Result<T, std::error::Error>;
```

Cargo.toml
```diff
@@ -50,6 +51,7 @@ restriction = "deny"
 nursery = "deny"
 cargo = "deny"
 
+allow-attributes-without-reason = { level = "allow", priority = 127 }
```
guzman-raphael: We shouldn't add an exception for this error. All clippy attributes added should include a reason/justification to document why it was necessary for that case.

src/util.rs
```diff
@@ -1,20 +1,16 @@
 use sha2::{Digest, Sha256};
 use std::any::type_name;
 
-#[expect(
-    clippy::unwrap_used,
-    reason = "`None` impossible since `type_name` always returns `&str`."
-)]
+#[expect(clippy::unwrap_used)]
```
guzman-raphael: See comment above. Reason should not have been removed.

Cargo.toml
```diff
@@ -72,6 +74,7 @@ missing_asserts_for_indexing = { level = "allow", priority = 127 }     # missing
 missing_docs_in_private_items = { level = "allow", priority = 127 }    # missing docs on private ok
 missing_inline_in_public_items = { level = "allow", priority = 127 }   # let rust compiler determine best inline logic
 missing_trait_methods = { level = "allow", priority = 127 }            # allow in favor of rustc `implement the missing item`
+multiple_crate_versions = { level = "allow", priority = 127 }
```
guzman-raphael: I've just tried removing this on my side and there was no error. Perhaps this change is now stale so we should remove it. Also, when adding a new one, we should use a comment to give context why this was necessary.

Cargo.toml
```diff
@@ -93,9 +96,11 @@ std_instead_of_core = { level = "allow", priority = 127 }              # we shou
 string_add = { level = "allow", priority = 127 }                       # simple concat ok
 string_lit_chars_any = { level = "allow", priority = 127 }             # favor readability until a perf case comes up
 todo = { level = "warn", priority = 127 }                              # warn todos
+verbose_file_reads = { level = "allow", priority = 127 }
```
guzman-raphael: I've just tried removing this on my side and there was no error. Perhaps this change is now stale so we should remove it. Also, when adding a new one, we should use a comment to give context why this was necessary.

Cargo.toml
```diff
@@ -93,9 +96,11 @@ std_instead_of_core = { level = "allow", priority = 127 }              # we shou
...
 use_debug = { level = "warn", priority = 127 }                         # debug print
-
+unseparated_literal_suffix = { level = "allow", priority = 127 }
```
guzman-raphael: I've just tried removing this on my side and there was no error. Perhaps this change is now stale so we should remove it. Also, when adding a new one, we should use a comment to give context why this was necessary.

Cargo.toml
```diff
@@ -93,9 +96,11 @@ std_instead_of_core = { level = "allow", priority = 127 }              # we shou
...
 # temporary
+missing_panics_doc = { level = "allow", priority = 127 }
```
guzman-raphael: I've just tried removing this on my side and there was no error. Perhaps this change is now stale so we should remove it. Also, when adding a new one, we should use a comment to give context why this was necessary.

src/store/mod.rs
```diff
@@ -1,11 +1,13 @@
...
-use std::{collections::BTreeMap, error::Error};
+use anyhow::Result;
+use std::collections::BTreeMap;
 
 pub trait Store {
-    fn save_pod(&self, pod: &Pod) -> Result<(), Box<dyn Error>>;
-    fn load_pod(&self, name: &str, version: &str) -> Result<Pod, Box<dyn Error>>;
-    fn list_pod(&self) -> Result<BTreeMap<String, Vec<String>>, Box<dyn Error>>;
-    fn delete_pod(&self, name: &str, version: &str) -> Result<(), Box<dyn Error>>;
+    fn save_pod(&mut self, pod: &Pod) -> Result<()>;
```
guzman-raphael: Why was it necessary to make `self` a `mut`? Because of our immutability design, we've generally found ways of not needing this. I'm missing some context/justification here.

src/store/mod.rs
```diff
@@ -1,11 +1,13 @@
...
+    fn load_pod(&mut self, name: &str, version: &str) -> Result<Pod>;
+    fn list_pod(&mut self) -> Result<&BTreeMap<String, String>>;
```
guzman-raphael: Unclear why the output type changed here. Keep in mind the idea was to use this output to display a table to the user (UI or CLI).

A typical programming pattern is to represent tables as as an ordered mapping where keys are the column name and value is an array of row values (all with the same length). This scales well, is easy to modify, and gives greater flexibility to the caller.

I think I remember you saying that the value might have become a concatenated value of 2 fields? If so, not sure I understand why it is better to always require string splitting of each row by the caller if they need to consume them.

src/store/mod.rs
```diff
@@ -1,11 +1,13 @@
...
+    fn delete_pod(&mut self, name: &str, version: &str) -> Result<()>;
```
guzman-raphael: Assuming `delete_annotation` below is the explicit delete of 1 annotation, would this then be an explicit delete of a whole pod i.e. the pod spec directory + all annotation references?

If so, then it might be better to change the input signature to just require a `hash` argument.

src/error.rs
```diff
@@ -1,5 +1,4 @@
 use colored::Colorize;
-use serde_yaml;
```
guzman-raphael: We discussed this at some point with @eywalker. It is isn't required by Rust since you can inline import, however, we mentioned following a good practice for ourselves of including everything that is coming externally at the top through `use` statements for clarity.

Unfortunately clippy doesn't have a lint for the ones that are only 1 level deep...

src/store/mod.rs
```diff
@@ -1,11 +1,13 @@
...
+use anyhow::Result;
+use std::collections::BTreeMap;
 
 pub trait Store {
-    fn save_pod(&self, pod: &Pod) -> Result<(), Box<dyn Error>>;
-    fn load_pod(&self, name: &str, version: &str) -> Result<Pod, Box<dyn Error>>;
-    fn list_pod(&self) -> Result<BTreeMap<String, Vec<String>>, Box<dyn Error>>;
-    fn delete_pod(&self, name: &str, version: &str) -> Result<(), Box<dyn Error>>;
+    fn save_pod(&mut self, pod: &Pod) -> Result<()>;
+    fn load_pod(&mut self, name: &str, version: &str) -> Result<Pod>;
```
guzman-raphael: This PR seems to also address making `Annotation` optional. If so, I would expect `load_pod` to handle loading with name/version (with annotation) or using hash (w/o annotation). The corresponding [issue](https://github.com/walkerlab/orcapod/issues/27) has more info on these cases.

My recommendation would actually be to move that feature into a separate PR since it is starting to become too much for one PR. That way it isn't all-or-nothing.

guzman-raphael (changes requested):
@Synicix Thanks for the updates. :handshake:

Submitting PR review progress for now but still need to review the following when I get back.

- src/model.rs
- src/filestore.rs
- tests/fixture/mod.rs
- tests/model.rs
- tests/store.rs

Cargo.toml
Synicix: Wouldn't that still require you to import Result from error now, when it is not really tied to that? In short anyhow is just a generic wrapper that allow for more stack tracing that takes any error type.

Cargo.toml
Synicix: I removed it mainly because of test that would require a lot of these exception when it is pretty obvious as to the reason why. Added it back in

Cargo.toml
Synicix: It had to do with fs4 and colorized. I remove fs4 for now since we are not doing file operations, but I can include the reason why

Cargo.toml
Synicix: Yeah this was an issue with fs4, I removed it because it was complaining that
``` rust
let file = fs::OpenOptions::new().read(true).open("foo.txt");
file.exclusive_lock()
```

should be
```
fs::read(<file>)
```

The issue is that, fs4 requires verbose file reads, and we will most likely run into it later when we start to use that. But for not I can remove it

Cargo.toml
Synicix: Odd, before than it was complaining about the regex. Guess when I rebuilt the container it might have became stale with the new standards. Removed.

Cargo.toml
Synicix: Issue with this was missing_panic_doc was triggering on the helper functions in test that do panic due to unwrap()
```
error: docs for function which may panic missing `# Panics` section
  --> tests/fixture/mod.rs:25:5
   |
25 |     pub fn get_name(&self) -> &str {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: first possible panic found here
  --> tests/fixture/mod.rs:27:32
   |
27 |             Self::Pod(pod) => &pod.annotation.as_ref().unwrap().name,
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_panics_doc
   = note: `-D clippy::missing-panics-doc` implied by `-D clippy::pedantic`
   = help: to override `-D clippy::pedantic` add `#[allow(clippy::missing_panics_doc)]`

error: docs for function which may panic missing `# Panics` section
  --> tests/fixture/mod.rs:38:5
   |
38 |     pub fn get_version(&self) -> &str {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: first possible panic found here
  --> tests/fixture/mod.rs:40:32
   |
40 |             Self::Pod(pod) => &pod.annotation.as_ref().unwrap().version,
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_panics_doc

error: docs for function which may panic missing `# Panics` section
  --> tests/fixture/mod.rs:51:5
   |
51 |     pub fn set_name(&mut self, name: &str) {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: first possible panic found here
  --> tests/fixture/mod.rs:53:52
   |
53 |             Self::Pod(pod) => name.clone_into(&mut pod.annotation.as_mut().unwrap().name),
   |                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_panics_doc
```

I added the reason for adding it

src/store/mod.rs
Synicix: We have to make it mut, due to self-containing the cache structure which needs to be updated in memory. The previous version stored it on the disk hence why we were able to get away with it.

Synicix: I think discussion can be continue on slack

src/store/mod.rs
Synicix: Currently it just forward back the cache which is comprise of BtreeString<(name-ver), (hash)> for easy look up and table construction.

I would be okay with converting where keys become the column name. What do you think @eywalker

src/store/mod.rs
Synicix: delete annotation, will only delete annotation and not the underlying object. Going off of what I remember we discussed last week. Delete_pod will wipe the entire item along with all annotations

src/error.rs
Synicix: I think the reason why I removed it was because it conflicts with Error thus we need to explicitly define it as serde_yaml::Error making the import moot. I can add it back in for the sake of clairty

src/store/mod.rs
Synicix: A few of the changes deals provides the ability to do that a bit easier in the future, but since we already decided on going with enum to identify it, I can code it in by tonight.

Synicix: Update pull request to have handle this case along with tests.

Synicix:
Removed cache for now @guzman-raphael @eywalker  Ready for review

src/model.rs
```diff
@@ -22,31 +17,31 @@ pub fn to_yaml<T: Serialize>(instance: &T) -> Result<String, Box<dyn Error>> {
...
 pub fn from_yaml<T: DeserializeOwned>(
-    annotation_file: &Path,
-    spec_file: &Path,
+    spec_yaml: &str,
     hash: &str,
-) -> Result<T, Box<dyn Error>> {
-    let annotation: Mapping = serde_yaml::from_str(&fs::read_to_string(annotation_file)?)?;
-    let spec_yaml = BufReader::new(fs::File::open(spec_file)?)
-        .lines()
-        .skip(1)
```
guzman-raphael, line 45: I see you removed the skip of the first line which had the `class` attribute. Is it not needed anymore? Is `serde_yaml` smart enough to see that in `T` (`Pod`) it isn't defined and therefore it skips it automatically?

src/model.rs
```diff
@@ -22,31 +17,31 @@ pub fn to_yaml<T: Serialize>(instance: &T) -> Result<String, Box<dyn Error>> {
...
-    Ok(instance)
+    Ok(serde_yaml::from_str(&serde_yaml::to_string(
+        &spec_mapping,
+    )?)?)
 }
 
 // --- core model structs ---
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
```
guzman-raphael: What was the reason you needed `PartialEq` and `Clone` on several of these? Was it for asserts in tests?

src/model.rs
```diff
@@ -71,9 +66,9 @@ impl Pod {
         recommended_cpus: f32,
         recommended_memory: u64,
         required_gpu: Option<GPURequirement>,
-    ) -> Result<Self, Box<dyn Error>> {
+    ) -> Result<Self> {
         let pod_no_hash = Self {
-            annotation,
+            annotation: Some(annotation),
```
guzman-raphael: Rather than doing this, I think it would be better to make `annotation` optional in `new`'s arguments as well so you can initialize a `pod` w/o annotation too. Then, it would should be a pass-through like the others.

src/store/filestore.rs
```diff
@@ -1,213 +1,216 @@
...
+
+            let Some(cap) = re.captures(&path_str) else {
+                continue;
+            };
+
+            // Insert into the cache
+            name_ver_tree.insert(
+                format!("{}-{}", &cap["name"].to_string(), &cap["ver"].to_string()),
+                cap["hash"].into(),
+            );
```
guzman-raphael: This still seems like a cache implementation builtin to the logic since it applies a copy.

src/store/filestore.rs
```diff
@@ -1,213 +1,216 @@
...
 pub struct LocalFileStore {
-    pub directory: PathBuf,
+    directory: PathBuf,
 }
 
 impl Store for LocalFileStore {
-    fn save_pod(&self, pod: &Pod) -> Result<(), Box<dyn Error>> {
-        let class = "pod";
+    fn save_pod(&self, pod: &Pod) -> Result<()> {
+        self.save_item(pod, &pod.hash, pod.annotation.as_ref())
```
guzman-raphael: `item` might not be clear/intuitive. We've been calling the core classes *models* which was borrowed from [Pydantic](https://docs.pydantic.dev/latest/concepts/models/) community. Perhaps we can rename `item` to `instance` similarly to how they refer to it?

src/store/filestore.rs
```diff
@@ -1,213 +1,216 @@
...
+    /// ```
+    fn save_item<T: Serialize>(
+        &self,
+        item: &T,
+        hash: &str,
+        annotation: Option<&Annotation>,
+    ) -> Result<()> {
+        // Save the item first
+        Self::save_file(
+            &self.make_path::<T>(hash, "spec.yaml"),
```
guzman-raphael: The intent behind `make_annotation_path`, `make_spec_path`, and `parse_annotation_path` was so that there was no other mention of paths anywhere so that any change to paths or to file naming would just affect those methods so it takes minimal work to make updates later.

It seems odd that we expect the caller to *know* the file is called `spec.yaml` which also breaks in more places if we change this later.

src/store/filestore.rs
```diff
@@ -1,213 +1,216 @@
...
+    }
 
-        // Save the pod and skip if it already exist, for the case of many annotation to a single pod
-        Self::save_file(
-            &self.make_spec_path(class, &pod.hash),
-            &to_yaml::<Pod>(pod)?,
-            false,
-        )?;
+    /// Return the name version index btree where key is name-version and value is hash of pod
+    fn list_pod(&self) -> Result<BTreeMap<String, String>> {
```
guzman-raphael: Similar to my previous feedback [here](https://github.com/walkerlab/orcapod/pull/34#discussion_r1812862407), I still prefer having tabular data represented as a `BTreeMap<String, Vec<_>>` where the key is column name and the vector is the values for the rows.

src/store/filestore.rs
```diff
@@ -1,213 +1,216 @@
...
         Ok(())
     }
 }
 
 impl LocalFileStore {
-    pub fn new(directory: impl Into<PathBuf>) -> Self {
+    pub fn new(directory: impl AsRef<Path>) -> Self {
         Self {
-            directory: directory.into(),
+            directory: directory.as_ref().into(),
```
guzman-raphael, line 64: Why is this better? Is there some documentation you could point me to?

Cargo.toml
guzman-raphael: We shouldn't rely on an external dependency when we can do this ourselves with just one line. Also, this will be changed anyway once #37 is merged.

src/store/filestore.rs
```diff
@@ -1,213 +1,216 @@
...
+        let re = Regex::new(&format!(
+            r"^.*\/{type_name}\/(?<hash>[a-z0-9]+)\/annotations\/(?<name>[A-z0-9\- ]+)-(?<ver>[0-9]+.[0-9]+.[0-9]+).yaml$"
+        ))?;
+
+        // Create the missing cache btree for the item
+        let mut name_ver_tree = BTreeMap::new();
+
+        let search_pattern = self.make_dir_path::<T>("*").join("annotations/*");
+
+        for path in glob::glob(&search_pattern.to_string_lossy())? {
```
guzman-raphael, line 190: In general, it is *usually* better to perform an iterator (lazy) approach as opposed to explicit loops. Here are a few reference articles for Rust specifically that can give some context where I am coming from.

- https://www.lurklurk.org/effective-rust/iterators.html
- https://doc.rust-lang.org/book/ch13-04-performance.html

guzman-raphael (changes requested):
Updating my review. Haven't gone through the tests much since I'll probably do a follow-up pass over it after merge anyway to help out.

src/model.rs
Synicix: Yes, if the field is not defined in the class it skips it and doesn't load it in.

Cargo.toml
Synicix: Perhaps I didn't do a good job of explaining it. The anyhow Error is a wrapper around the std::error that has some additional feature regarding backtrace. It doesn't really change the behavior but instead add on to it.

More info can be found here: https://docs.rs/anyhow/latest/anyhow/

src/model.rs
Synicix: Fair, since we are going to use the none annotation later on anyways. Will change

src/store/filestore.rs
Synicix: Not sure instance is a good name for this either since isn' that also pretty generic too?

src/store/filestore.rs
Synicix: One problem with this is the vectors will be locked a specific type such as String. I assume we are okay with that? A vector of struct would be more flexiable in the the future instead of a btreemap

src/store/filestore.rs
Synicix: We don't really want to force it to covert into PathBuf, instead we just want it to implment the Path reference.
Pathbuf is like String and path is like &str. We got the idea from how std::path lib implments it: https://doc.rust-lang.org/src/std/path.rs.html#3262

src/store/filestore.rs
Synicix: I prefer readability in this case since I didn't want to redefine the regex pattern every loop like you had in your previous example. Plus it had that odd return type.

src/store/filestore.rs
Synicix: The comment is stale, it still uses the same structure as before, just without storing it in memory. The Btreemap is just a nice way to quickly look up annotation-version -> hash. I will update the comment

Synicix:
@eywalker @guzman-raphael Ready for review again, but probably need to discuss on some of the points above.

src/error.rs
```diff
@@ -8,10 +9,9 @@ use std::{
     fmt::{Display, Formatter},
     io,
     path::PathBuf,
-    result,
 };
 /// Shorthand for a Result that returns an `OrcaError`.
-pub(crate) type Result<T> = result::Result<T, OrcaError>;
+pub type Result<T> = anyhow::Result<T, OrcaError>;
```
eywalker, line 14: Let's discuss the use of `anyhow` crate along with whether to make this alias public.

src/error.rs
```diff
@@ -60,12 +60,17 @@ impl Display for OrcaError {
             Kind::NoAnnotationFound(class, name, version) => {
                 write!(f, "No annotation found for `{name}:{version}` {class}.")
             }
-            Kind::NoRegexMatch => {
```
eywalker, line 63: Why was this removed?

src/store/filestore.rs
eywalker: we could just use the term `model` no?

src/store/filestore.rs
eywalker: Yeah this allows for the method to more broadly accept different input types. It is a bit more idiomatic to use the following: PathBuf::from(directory.as_ref())

src/store/filestore.rs
eywalker: Let's put this as a point for a quick discussion

eywalker (changes requested):

src/error.rs
Synicix: Does regex function I rewrote allow for no mismatch and relies on it. So instead of erroring out it just returns and empty Btree

Cargo.toml
guzman-raphael: @Synicix is looking into why the alias isn't working when set in the library.

src/store/filestore.rs
guzman-raphael: Discussed with @Synicix. Planning to go with `item` -> `model` and `ItemKey` -> `ModelID`.

src/store/filestore.rs
```diff
@@ -1,207 +1,256 @@
...
+    /// this will break...
+    #[expect(
+        clippy::unwrap_in_result,
+        reason = "It shouldn't failed as we added the keys ourselves at the start"
+    )]
+    #[expect(
+        clippy::unwrap_used,
+        reason = "It shouldn't failed as we added the keys ourselves at the start"
+    )]
+    fn list_pod(&self) -> Result<BTreeMap<String, Vec<String>>> {
```
guzman-raphael: @Synicix Is going to change the return type to a `Vec<Struct>` and we'll revisit once the update is in.

src/store/filestore.rs
guzman-raphael: @guzman-raphael Will look into a few cases and report back.

src/store/filestore.rs
eywalker: Sounds good!

src/store/filestore.rs
Synicix: Done

src/model.rs
Synicix: Yep that is the main reason

src/store/filestore.rs
Synicix: Updated with static const now

src/store/filestore.rs
Synicix: Chhange to vector of struct

Synicix:
@guzman-raphael @eywalker So with the discussion that Rapahel and me had. I resolve some of the change requested with a few remaining.

- Unable to get the anyhow alias to work due to the fact that the return error type must be of anyhow::Error for the stack trace to work. Thus it has to be used on the top level for it to display to force conversion of our OrcaError into anyhow::Error
- Waiting for Raphael to see if there is a better implemnetation with maps regarding the name_ver_tree
- Test still doesn't do the fixture for the pod yet (might be able to pick this up later tonight or tomrrow)

src/error.rs
eywalker: I'm going to let this pass for now and we can decide to revisit the use of `anyhow` in the future PR. Fortunately, the usage pattern of `Result<T>` won't change.

src/store/filestore.rs
eywalker: In the interest of having some closure on this PR that's been up for a while, let's set this as a separate issue and address in the future.

src/store/mod.rs
```diff
@@ -1,5 +1,22 @@
 use crate::{error::Result, model::Pod};
-use std::collections::BTreeMap;
+
+/// Enum for identification to
+pub enum ModelID {
```
eywalker, line 4: We should consider making this clonable

src/store/mod.rs
```diff
@@ -15,20 +32,26 @@ pub trait Store {
...
     /// `version`.
-    fn load_pod(&self, name: &str, version: &str) -> Result<Pod>;
+    fn load_pod(&self, model_id: &ModelID) -> Result<Pod>;
     /// How to query stored pods.
     ///
     /// # Errors
     ///
     /// Will return `Err` if there is an issue querying metadata from existing pods in the store.
-    fn list_pod(&self) -> Result<BTreeMap<String, Vec<String>>>;
+    fn list_pod(&self) -> Result<Vec<ModelInfo>>;
```
eywalker, line 41: Let's follow up on the specification of list behavior as a dedicated issue.

eywalker (approved):
There are a few hanging points that I'd want to move as issues to be followed up on so that this PR can be merged. This branch has already been serving as the base for many development and it should be merged.

guzman-raphael (approved):
Merging for now to unblock and address the remaining points in a follow-up PR.

PR #35, opened 2024-10-17
Pod job

Synicix:
This will depend on #34 as it uses a lot of the generics I added from there

Until then a good preview of what changed form #34 is: https://github.com/Synicix/orcapod/compare/parse_annotation...Synicix:orcapod:pod-job?expand=1


Synicix:
Update it to follow new design, ready for review once the other pull request is merge

PR #37, opened 2024-10-21
Improve error handling + docs

### LINKED ISSUE #20: Add API docs
- [ ] Agree on API documentation style and standard
- [ ] Add reference API documentation on methods of `Filestore`

### LINKED ISSUE #25: Standarize error handling
e.g. based on an `OrcaError` enum

These seem like interesting options:
- https://stackoverflow.com/questions/48430836/rust-proper-error-handling-auto-convert-from-one-error-type-to-another-with-que
- https://doc.rust-lang.org/rust-by-example/error/multiple_error_types/wrap_error.html

guzman-raphael:
- [x] ~Convert error handling to enumeration approach~ Convert error handling to masked enumeration approach
- [x] Apply similar message highlighting as before
- [x] Add API docs
- [x] Make return type for `parse_annotation_path` more readable
- [x] In `save_file`, skip creating directory path as opposed to raising error if no parent
- [x] Improve clippy reason doc in `get_type_name`
- [x] Reduce scope of `expect_used` clippy exception
- [x] Remove temporary clippy exceptions for docs
- [x] Add notes about generating docs HTML site and coverage report HTML site

Fix #25 
Fix #20

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,4 @@
+use colored::Colorize;
 use glob;
 use regex;
 use serde_yaml;
@@ -61,10 +62,18 @@ impl Display for OrcaError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match &self.0 {
             Kind::FileExists(path) => {
-                write!(f, "File `{}` already exists.", path.to_string_lossy())
+                write!(
+                    f,
+                    "File `{}` already exists.",
+                    path.to_string_lossy().bright_cyan()
+                )
             }
             Kind::FileHasNoParent(path) => {
-                write!(f, "File `{}` has no parent.", path.to_string_lossy())
+                write!(
+                    f,
+                    "File `{}` has no parent.",
+                    path.to_string_lossy().bright_red()
+                )
             }
             Kind::NoAnnotationFound(class, name, version) => {
                 write!(f, "No annotation found for `{name}:{version}` {class}.")
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,9 +7,10 @@ use std::{
     fmt::{Display, Formatter},
     io,
     path::PathBuf,
+    result,
 };
 /// Shorthand for a Result that returns an `OrcaError`.
-pub type OrcaResult<T> = Result<T, OrcaError>;
+pub type Result<T> = result::Result<T, OrcaError>;
 /// Possiable errors you may encounter.
 #[derive(Debug)]
 pub enum OrcaError {
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,9 +11,10 @@ use std::{
 };
 /// Shorthand for a Result that returns an `OrcaError`.
 pub type Result<T> = result::Result<T, OrcaError>;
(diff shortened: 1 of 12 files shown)


src/error.rs
```diff
@@ -7,123 +8,76 @@ use std::{
     io,
     path::PathBuf,
 };
-
-/// Wrapper around `serde_yaml::from_str`
+/// Shorthand for a Result that returns an `OrcaError`.
+pub type OrcaResult<T> = Result<T, OrcaError>;
```
eywalker: I'd actually prefer we just use `Result` and `Error` instead of `OrcaResult` and `OrcaError` following other crates' practices and also because I also do find `orcapod:error:Error` is enough of a specifier

src/error.rs
```diff
@@ -7,123 +8,76 @@ use std::{
...
+/// Possiable errors you may encounter.
 #[derive(Debug)]
-pub struct DeserializeFailure {
-    pub path: PathBuf,
-    pub error: serde_yaml::Error,
+pub enum OrcaError {
```
eywalker: In libraries like serde_yaml and std::io, it seems to be a common practice to have a rather simple `Error` struct that implement std::error and have the `enum` type be held internally. Do you know why they do this indirection?

src/error.rs
eywalker: I also see this indirect `Error` struct to error Enum relationship found in library like `tokio` (https://docs.rs/tokio/latest/src/tokio/time/error.rs.html#26). This makes me believe that there is some decent reason to have this separation.

src/error.rs
eywalker: I also see this indirect Error struct to error Enum relationship found in library like tokio (https://docs.rs/tokio/latest/src/tokio/time/error.rs.html#26). This makes me believe that there is some decent reason to have this separation.

src/error.rs
eywalker: Actually, I do see that more modern recommendation is to give more specific name like you did. Do you happen to have a good reference on what's a good practices on error implementation these days?

src/error.rs
eywalker: Here is an explanation for why this separation is recommended:

----
This is a common pattern in Rust for error handling that offers several benefits. Let me explain the key advantages:

1. Encapsulation and API Stability: 
  * By wrapping the Kind enum in an Error struct, the internal representation (Kind) can be changed without breaking public API compatibility
  * The pub(crate) on Kind means it's only visible within the crate, while Error is public
  * This allows the library to add new error kinds or restructure the internal representation without requiring users to update their code


2 Controlled Error Creation:
  * The Error struct provides specific constructor methods (shutdown(), at_capacity(), invalid())
  * This gives more control over how errors are created and prevents users from directly creating invalid error states
  * The associated methods like is_shutdown() provide a clean API for error checking


3 Pattern Matching Protection:
  * If Kind was directly exposed, users might pattern match on it, making their code brittle to changes
  * By hiding Kind, the library ensures users rely on the provided methods instead

Synicix (approved):
Looks good merging

PR #38, opened 2024-10-29
Share `Kind` within crate and create `OrcaError` from `Kind`

guzman-raphael:


Synicix (approved):
Looks good

src/error.rs
```diff
@@ -11,11 +11,11 @@ use std::{
     result,
 };
 /// Shorthand for a Result that returns an `OrcaError`.
-pub type Result<T> = result::Result<T, OrcaError>;
+pub(crate) type Result<T> = result::Result<T, OrcaError>;
```
eywalker, line 14: Given we will have methods that will actually return this Result<T>, wouldn't the unavailability of this type definition make the code uncompilable the moment those methods using Result<T> return types are used outside of our crate? Does it act differently if this is a type alias?

tests/fixture/mod.rs
```diff
@@ -4,13 +4,15 @@
...
-    error::Result,
+    error::OrcaError,
     model::{Annotation, Pod, StreamInfo},
     store::{filestore::LocalFileStore, Store},
 };
-use std::{collections::BTreeMap, fs, ops::Deref, path::PathBuf};
+use std::{collections::BTreeMap, fs, ops::Deref, path::PathBuf, result};
 use tempfile::tempdir;
 
+pub type Result<T> = result::Result<T, OrcaError>;
```
eywalker, line 14: Given users might also want to use our alias for Result if they work a lot with OrcaError, we might as well treat it as publicly facing interface?

eywalker (commented):
I believe there might be a problem with the handling of the `Result<T>` alias?

eywalker:
After further digging, I learned that it is not necessary for type def to be shared publicly even if used as a return type for a public method as it is just a convenience alias available within the crate (if you so choose). So the concern over the code not compiling was not well founded - sorry! That being said, should we still consider making the alias publicly available? I see it's fairly common to make it public.

PR #42, opened 2024-11-01
Generics and annotation improvements, Part 2

### LINKED ISSUE #26: Make `Store` trait methods generic

### LINKED ISSUE #27: Make `Annotation` optional

### LINKED ISSUE #36: Change Annotation handling

### LINKED ISSUE #39: Change regex to mapping function
Raphael might have something that is better than explicit for loops

guzman-raphael:
Following the merge of #34 , this PR applies the ideas in my review.

Below is a summary of all the features that were accomplished across both PRs.

## Features

- Use generic methods
- Make deletes explicit (`delete_annotation`, `delete_pod`), remove checks in `delete_pod` since deletes are explicit, and prevent deletion of the last annotation in `delete_annotation`
- Change annotation save path (`annotation` subdirectory in model directory)
- Make annotation optional for `load_pod` and `delete_pod` but required for `save_pod`
- Support loading or deleting pod from a hash or name/version (via `ModelID`)
- Make `Result` pub
- Expose `LocalFileStore.directory` only via getter
- Use `impl AsRef<Path>` as opposed to `impl Into<PathBuf>`
- Apply map to format desired in `list_model`. Currently, just passing through `Vec<ModelInfo>`.

## Minor Improvements

- Make path builder more general
- Use a `Vec<ModelInfo>` internally when parsing annotation path
- Compile regex once instead of in each iteration (moved out of map via [move](https://doc.rust-lang.org/std/keyword.move.html))
- Parse `class` in regex
- Make `save_file` more flexible
- Use YAML content as opposed to files in `from_yaml`
- Add `lookup_hash` based on `name`/`version`
- Remove `get_version_map` since unneeded with explicit deletes
- Remove `FileHasNoParent` error
- Rename `parse_annotation_path(path)` -> `find_annotation(glob_pattern)` to improve readability.
- Add `snake_case` utility
- Simplify util methods
- Simplify `to_yaml` usage
- Update docs

## Tests

- ~Removed tests introduced in #34 so they are addressed separately~ Tests have been reworked to align with the original fixture design. Includes tests from #34.  Improves code coverage from ~70% -> ~72%.
- Allow `expect` usage in tests to simplify
- Add `pod_load` test
- Add equality support in `Pod`
- Add test for `list_pod` when none exist yet
- Create store dir in fixture when initializing so drop doesn't panic
- Make empty dir check in tests take an arg for # of levels up
- Pull out implementations from fixture functions

## Dev Improvements

- Make spec/annotation paths and filenames easier to change in the future
- Add spell checker w/ a word ignore list
- Fix typos

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,3 @@
-use anyhow;
 use colored::Colorize;
 use glob;
 use regex;
@@ -9,24 +8,28 @@ use std::{
     fmt::{Display, Formatter},
     io,
     path::PathBuf,
+    result,
 };
 /// Shorthand for a Result that returns an `OrcaError`.
-pub type Result<T> = anyhow::Result<T, OrcaError>;
-// pub type Result<T> = result::Result<T, OrcaError>;
+pub type Result<T> = result::Result<T, OrcaError>;
 
 /// Possible errors you may encounter.
 #[derive(Debug)]
 pub(crate) enum Kind {
     /// Returned if a file is not expected to exist.
     FileExists(PathBuf),
-    /// Returned if a file is expected to have a parent.
-    FileHasNoParent(PathBuf),
     /// Returned if an annotation was expected to exist.
     NoAnnotationFound(String, String, String),
+    /// Returned if a model save was attempted without an annotation set.
+    MissingAnnotationOnSave,
+    /// Returned if an annotation delete was attempted on a model's last annotation.
+    DeletingLastAnnotation(String, String, String),
+    /// Returned if a regular expression was expected to match.
+    NoRegexMatch,
     /// Wrapper around `glob::GlobError`
     GlobError(glob::GlobError),
     /// Wrapper around `glob::PatternError`
-    GlobPaternError(glob::PatternError),
+    GlobPatternError(glob::PatternError),
     /// Wrapper around `regex::Error`
     RegexError(regex::Error),
     /// Wrapper around `serde_yaml::Error`
@@ -49,18 +52,23 @@ impl Display for OrcaError {
                     path.to_string_lossy().bright_cyan()
                 )
             }
-            Kind::FileHasNoParent(path) => {
-                write!(
-                    f,
-                    "File `{}` has no parent.",
-                    path.to_string_lossy().bright_red()
-                )
-            }
             Kind::NoAnnotationFound(class, name, version) => {
                 write!(f, "No annotation found for `{name}:{version}` {class}.")
             }
+            Kind::MissingAnnotationOnSave => {
+                write!(f, "No annotation found when attempting to store.")
+            }
+            Kind::DeletingLastAnnotation(class, name, version) => {
+                write!(
+                    f,
+                    "Attempted to delete the last annotation for `{name}:{version}` {class}."
+                )
+            }
+            Kind::NoRegexMatch => {
+                write!(f, "No match for regex.")
+            }
             Kind::GlobError(error) => write!(f, "{error}"),
-            Kind::GlobPaternError(error) => write!(f, "{error}"),
+            Kind::GlobPatternError(error) => write!(f, "{error}"),
             Kind::SerdeYamlError(error) => write!(f, "{error}"),
             Kind::RegexError(error) => write!(f, "{error}"),
             Kind::IoError(error) => write!(f, "{error}"),
@@ -74,7 +82,7 @@ impl From<glob::GlobError> for OrcaError {
 }
 impl From<glob::PatternError> for OrcaError {
     fn from(error: glob::PatternError) -> Self {
-        Self(Kind::GlobPaternError(error))
+        Self(Kind::GlobPatternError(error))
     }
 }
 impl From<serde_yaml::Error> for OrcaError {
--- a/src/error.rs
+++ b/src/error.rs
@@ -42,6 +42,12 @@ pub(crate) enum Kind {
 #[derive(Debug)]
 pub struct OrcaError(Kind);
 impl Error for OrcaError {}
+impl OrcaError {
+    /// Returns `true` if the error was caused by an attempt to delete a model's last annotation.
+    pub const fn is_deleting_last_annotation(&self) -> bool {
+        matches!(self.0, Kind::DeletingLastAnnotation(_, _, _))
+    }
+}
 impl Display for OrcaError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match &self.0 {
--- a/src/model.rs
+++ b/src/model.rs
@@ -29,31 +29,25 @@ pub fn to_yaml<T: Serialize>(instance: &T) -> Result<String> {
 /// Will return `Err` if there is an issue converting YAML files for spec+annotation into a model
 /// instance.
 pub fn from_yaml<T: DeserializeOwned>(
-    spec_yaml: &str,
     hash: &str,
+    spec_yaml: &str,
     annotation_yaml: Option<&str>,
 ) -> Result<T> {
-    let mut spec_mapping: BTreeMap<String, Value> = serde_yaml::from_str(spec_yaml)?;
-
-    // Insert annotation if there is something
-    if let Some(yaml) = annotation_yaml {
-        let annotation_map: Mapping = serde_yaml::from_str(yaml)?;
-        spec_mapping.insert("annotation".into(), Value::from(annotation_map));
+    let mut spec: BTreeMap<String, Value> = serde_yaml::from_str(spec_yaml)?;
+    spec.insert("hash".to_owned(), Value::from(hash));
+    if let Some(resolved_annotation_yaml) = annotation_yaml {
+        let annotation: Mapping = serde_yaml::from_str(resolved_annotation_yaml)?;
+        spec.insert("annotation".to_owned(), Value::from(annotation));
     }
-    spec_mapping.insert("hash".to_owned(), Value::from(hash));
 
-    Ok(serde_yaml::from_str(&serde_yaml::to_string(
-        &spec_mapping,
-    )?)?)
+    Ok(serde_yaml::from_str(&serde_yaml::to_string(&spec)?)?)
 }
 
 // --- core model structs ---
 
 /// A reusable, containerized computational unit.
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Pod {
-    /// Metadata that doesn't affect reproducibility.
-    pub annotation: Option<Annotation>,
     /// Unique id based on reproducibility.
     pub hash: String,
     source_commit_url: String,
@@ -64,6 +58,8 @@ pub struct Pod {
     output_stream_map: BTreeMap<String, StreamInfo>,
     recommended_cpus: f32,
     recommended_memory: u64,
+    /// Metadata that doesn't affect reproducibility.
+    pub annotation: Option<Annotation>,
     required_gpu: Option<GPURequirement>,
 }
 
@@ -74,7 +70,6 @@ impl Pod {
     ///
     /// Will return `Err` if there is an issue initializing a `Pod` instance.
     pub fn new(
-        annotation: Option<Annotation>,
         source_commit_url: String,
         image: String,
         command: String,
@@ -83,10 +78,10 @@ impl Pod {
         output_stream_map: BTreeMap<String, StreamInfo>,
         recommended_cpus: f32,
         recommended_memory: u64,
+        annotation: Option<Annotation>,
         required_gpu: Option<GPURequirement>,
     ) -> Result<Self> {
         let pod_no_hash = Self {
-            annotation,
             hash: String::new(),
             source_commit_url,
             image,
@@ -96,6 +91,7 @@ impl Pod {
             output_stream_map,
             recommended_cpus,
             recommended_memory,
+            annotation,
             required_gpu,
         };
         Ok(Self {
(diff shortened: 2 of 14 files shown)


src/error.rs
```diff
@@ -9,24 +8,28 @@ use std::{
...
     /// Returned if a file is not expected to exist.
     FileExists(PathBuf),
-    /// Returned if a file is expected to have a parent.
-    FileHasNoParent(PathBuf),
     /// Returned if an annotation was expected to exist.
     NoAnnotationFound(String, String, String),
+    /// Returned if a model save was attempted without an annotation set.
+    MissingAnnotationOnSave,
+    /// Returned if an annotation delete was attempted on a model's last annotation.
+    DeletingLastAnnotation(String, String, String),
```
Synicix, line 26: Don't we want to support annotationless models?

src/model.rs
```diff
@@ -29,31 +29,25 @@ pub fn to_yaml<T: Serialize>(instance: &T) -> Result<String> {
...
     annotation_yaml: Option<&str>,
 ) -> Result<T> {
-    let mut spec_mapping: BTreeMap<String, Value> = serde_yaml::from_str(spec_yaml)?;
-
-    // Insert annotation if there is something
-    if let Some(yaml) = annotation_yaml {
-        let annotation_map: Mapping = serde_yaml::from_str(yaml)?;
-        spec_mapping.insert("annotation".into(), Value::from(annotation_map));
+    let mut spec: BTreeMap<String, Value> = serde_yaml::from_str(spec_yaml)?;
+    spec.insert("hash".to_owned(), Value::from(hash));
```
Synicix, line 37: We storing the hash now?

src/model.rs
```diff
@@ -29,31 +29,25 @@ pub fn to_yaml<T: Serialize>(instance: &T) -> Result<String> {
...
-    Ok(serde_yaml::from_str(&serde_yaml::to_string(
-        &spec_mapping,
-    )?)?)
+    Ok(serde_yaml::from_str(&serde_yaml::to_string(&spec)?)?)
 }
 
 // --- core model structs ---
 
 /// A reusable, containerized computational unit.
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
```
Synicix, line 53: I would probably leave clone there because we will need it later chances are

src/model.rs
```diff
@@ -64,6 +58,8 @@ pub struct Pod {
     output_stream_map: BTreeMap<String, StreamInfo>,
     recommended_cpus: f32,
     recommended_memory: u64,
+    /// Metadata that doesn't affect reproducibility.
+    pub annotation: Option<Annotation>,
```
Synicix: I am not so sure of moving annotation to a lower line

src/model.rs
```diff
@@ -83,10 +78,10 @@ impl Pod {
         output_stream_map: BTreeMap<String, StreamInfo>,
         recommended_cpus: f32,
         recommended_memory: u64,
+        annotation: Option<Annotation>,
```
Synicix: Similar to above, I feel like when defining the pod, the name description and version comes to mind first follow by the rest of the spec?

src/model.rs
```diff
@@ -108,7 +104,7 @@ impl Pod {
 // --- util types ---
 
 /// Standard metadata structure for all model instances.
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
Synicix, line 111: Perhaps leave Clone in there

src/store/filestore.rs
```diff
@@ -11,36 +12,29 @@ use std::{
...
 #[derive(Debug)]
 pub struct LocalFileStore {
+    /// A local path to a directory where store will be located.
     directory: PathBuf,
 }
 
 impl Store for LocalFileStore {
     fn save_pod(&self, pod: &Pod) -> Result<()> {
-        self.save_model(pod, &pod.hash, pod.annotation.as_ref())
+        self.save_model(
```
Synicix, line 23: This doesn't allow saving without an annotation, making annotations more of a requirement. Perhaps we need to discuss and decide on how optional is annotation

src/store/filestore.rs
```diff
@@ -11,36 +12,29 @@ use std::{
...
     fn load_pod(&self, model_id: &ModelID) -> Result<Pod> {
-        self.load_model::<Pod>(model_id)
+        self.load_model(model_id)
     }
 
-    /// Return Btree where key is column name and value is a vec of values
-    /// Are we okay with returning a bunch of strings? For now it works but later on like adding version
-    /// this will break...
-    fn list_pod(&self) -> Result<Vec<ModelInfo>> {
+    fn list_pod(&self) -> Result<BTreeMap<String, Vec<String>>> {
```
Synicix: Thought we agree to change it to a struct?

src/store/filestore.rs
```diff
@@ -49,206 +43,196 @@ impl Store for LocalFileStore {
...
             "{}/{}/{}",
             self.directory.to_string_lossy(),
             get_type_name::<T>(),
-            hash,
+            hash
         ))
+        .join(relpath)
+    }
+
+    fn parse_annotation_path(path: &Path) -> Result<impl Iterator<Item = Result<ModelInfo>>> {
```
Synicix: Perhaps instead of path, do glob search pattern because that more accurately describes what it is?

src/store/filestore.rs
```diff
@@ -49,206 +43,196 @@ impl Store for LocalFileStore {
...
-    /// Helper function for making path to a given item type T
-    pub fn make_path<T>(&self, hash: &str, file_name: &str) -> PathBuf {
-        self.make_dir_path::<T>(hash).join(file_name)
+    fn lookup_hash<T>(&self, name: &str, version: &str) -> Result<String> {
+        let model_info = Self::parse_annotation_path(
+            &self.make_path::<T>("*", &Self::make_annotation_relpath(name, version)),
+        )?
+        .next()
+        .ok_or_else(|| {
+            OrcaError::from(Kind::NoAnnotationFound(
```
Synicix, line 127: This relates to the point of whether we want no annotation model to be possible

src/store/filestore.rs
```diff
@@ -49,206 +43,196 @@ impl Store for LocalFileStore {
...
     fn delete_annotation<T>(&self, name: &str, version: &str) -> Result<()> {
-        // Search the name ver index for the hash
-        let hash = self.get_hash_from_name_ver_tree::<T>(name, version)?;
+        let hash = self.lookup_hash::<T>(name, version)?;
+        let (count, _) = Self::parse_annotation_path(
+            &self.make_path::<T>(&hash, Self::make_annotation_relpath("*", "*")),
+        )?
+        .size_hint();
+        if count == 1 {
+            return Err(OrcaError::from(Kind::DeletingLastAnnotation(
```
Synicix, line 51: Same thing as above about the point relating to annoationless models

src/store/filestore.rs
```diff
@@ -49,206 +43,196 @@ impl Store for LocalFileStore {
...
-            ModelID::NameVer(name, version) => {
-                // Search the name-ver index
-                let hash = self.get_hash_from_name_ver_tree::<T>(name, version)?;
-
-                // Get the spec and annotation yaml
-                let spec_yaml = fs::read_to_string(self.make_path::<T>(&hash, SPEC_FILE_NAME))?;
-
-                let annotation_yaml =
-                    fs::read_to_string(self.make_annotation_path::<T>(&hash, name, version))?;
+    fn list_model<T>(&self) -> Result<BTreeMap<String, Vec<String>>> {
```
Synicix: Yeah we need to decide on what we want to send back for list

Synicix (changes requested):
So there is a couple of discussion points we should probably go over next time we meet:
- How to deal with annotationless models. Seems that we have different apporches on how to handle that
- Test changes should adopt new design for generics?
- What should list_model return

src/error.rs
guzman-raphael: Yes, we should but my question is whether `save_model` needs to support saving a model w/o annotation?

It came up while I was implementing this so thought I'd offer a solution for us to review. The reason I went with this approach was because it was easier to make it 'feature complete'.

If we allow saving a model w/o annotation, that annotation-less model now needs to be included in the return of `list_model`. Solving that will make the logic for `list_model` more complex/messier (w/o listing the first annotation twice).

src/model.rs
guzman-raphael: No, this line is in `from_yaml` so it is about creating a model in memory from YAML strings.

The main reason it shows up as a diff is because I moved the pieces about `spec` to be together instead of spliting it by `annotation` and renamed the variables to simplify. `spec_mapping` on a type of `BTreeMap` was a redundant name.

src/model.rs
guzman-raphael: You are right but I'd rather add it with the feature that requires it.

For example, if implementing `PodJob` requires this, then it would make sense to me to add that in the PR that introduces `PodJob` so that the diff reflects the need for it and its application. That way, it is less of a hassle to chase down what actually required this.

The way to interpret it here is that at a minimum, we don't need it for now with our current tests.

src/model.rs
guzman-raphael: I was trying to follow a common pattern of pushing optional args (ones with a default (`None`)) towards the end of constructors (`new`).

This is quite minor and I don't feel super strongly either way. Mainly just think we should align in general so we are consistent.

src/model.rs
guzman-raphael: See above [comment](https://github.com/walkerlab/orcapod/pull/42#discussion_r1827672438). This would be a good argument for pushing it higher and having user flow > putting optional args at the end.

src/model.rs
guzman-raphael: See my other [comment](https://github.com/walkerlab/orcapod/pull/42#discussion_r1827663606) on `Clone`.

src/store/filestore.rs
guzman-raphael: I agree. I implemented this way to open up discussion around this. Sometimes it is easier to see and have something in front of you to help the discussions. See my [comment](https://github.com/walkerlab/orcapod/pull/42#discussion_r1827652408) here for some context.

src/store/filestore.rs
guzman-raphael: We did and it has been implemented using a `Vec<ModelInfo>` *internally*.

I think it is important to distinguish between the structure it is parsed internally and the structure that returns to the user in this case. They could be the same but doesn't necessarily have to be. One is optimized for flexibility/traversal/lookups while the other is optimized for displaying the content in a table-friendly format. 

In general, I like the approach that there is a way to map the internal structure to a more convenience structure (depending on the caller's needs) without having a penalty of looping over the data multiple times.

src/store/filestore.rs
guzman-raphael: You are right and I like your suggestion. Updated it to `fn find_annotation(glob_pattern: &Path)`, what do you think?

src/store/filestore.rs
guzman-raphael: Actually, this is a bit different. This case happens when a user has tried to either `delete_model`, `delete_annotation`, or `load_model` specifically with the `Annotation` option but the `(name, version)` they supplied does not exist in the store i.e. they provided invalid user input.

src/store/filestore.rs
guzman-raphael: This one is related to requiring annotation on `save_model` and consequently, preventing the last annotation from being deleted so it won't *disappear* from `list_model` output. See my other [comment](https://github.com/walkerlab/orcapod/pull/42#discussion_r1827652408) for context.

src/store/filestore.rs
guzman-raphael: Yes, we should discuss. Left some thoughts in this [comment](https://github.com/walkerlab/orcapod/pull/42#discussion_r1827683220).

src/store/filestore.rs
Synicix: @guzman-raphael Change it back to model_info

guzman-raphael: Update has been pushed.

src/store/filestore.rs
guzman-raphael: From our discussion, I've updated it to just pass through `Vec<ModelInfo>` for now. We can revisit once we are closer to the UI display rendering. Should be very easy to adjust structure to whatever we need.

Synicix (approved):

PR #47, opened 2025-01-14
Pod Run and Pod Output with test

Synicix:
need to be heavily updated to the new design


Synicix:
deprecated, will rebased later on

PR #64, opened 2025-02-25
Add Crypto Module use for hashing

Synicix:
- Add hashing algorithm for chunk hashing #45 
- Update gpu docker dev container with additional flags to run the docker in docker tests
- Add some temporary dead code tolerations

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -20,6 +20,8 @@
 	},
 	"runArgs": [
 		"--name=${localWorkspaceFolderBasename}_devcontainer",
+		"--cpus=2",
+		"--memory=8gb",
 		"--gpus=all",
 		"--privileged",
 		"--cgroupns=host"
@@ -32,7 +34,9 @@
 				"ms-vscode.hexeditor", // Binary preview in HEX
 				"tamasfe.even-better-toml", // *.toml language support
 				"eamodio.gitlens", // Git explorer in VSCode
-				"streetsidesoftware.code-spell-checker" // Catch spelling errors in docs
+				"streetsidesoftware.code-spell-checker", // Catch spelling errors in docs
+				"GitHub.copilot-chat", // GitHub Copilot AI assistant
+				"tintinweb.graphviz-interactive-preview" // Graphviz DOT preview
 			]
 		}
 	}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,7 +33,9 @@
 				"ms-vscode.hexeditor", // Binary preview in HEX
 				"tamasfe.even-better-toml", // *.toml language support
 				"eamodio.gitlens", // Git explorer in VSCode
-				"streetsidesoftware.code-spell-checker" // Catch spelling errors in docs
+				"streetsidesoftware.code-spell-checker", // Catch spelling errors in docs
+				"GitHub.copilot-chat", // GitHub Copilot AI assistant
+				"tintinweb.graphviz-interactive-preview" // Graphviz DOT preview
 			]
 		}
 	}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-	"name": "Development",
+	"name": "Development (GPU)",
 	"build": {
 		"dockerfile": "Dockerfile"
 	},
@@ -12,7 +12,7 @@
 		"LOCAL_WORKSPACE_FOLDER": "${localWorkspaceFolder}",
 		"RUST_BACKTRACE": "1"
 	},
-	"postStartCommand": "docker system prune -fa && docker volume prune -f",
+	"postStartCommand": "docker system prune -fa && docker volume prune -f && apt update && apt install fish -y && chsh -s /usr/bin/fish",
 	"hostRequirements": {
 		"cpus": 2,
 		"memory": "8gb",
@@ -20,8 +20,7 @@
 	},
 	"runArgs": [
 		"--name=${localWorkspaceFolderBasename}_devcontainer",
-		"--cpus=2",
-		"--memory=8gb",
+		"--gpus=all",
 		"--privileged",
 		"--cgroupns=host"
 	],
@@ -33,9 +32,7 @@
 				"ms-vscode.hexeditor", // Binary preview in HEX
 				"tamasfe.even-better-toml", // *.toml language support
 				"eamodio.gitlens", // Git explorer in VSCode
-				"streetsidesoftware.code-spell-checker", // Catch spelling errors in docs
-				"GitHub.copilot-chat", // GitHub Copilot AI assistant
-				"tintinweb.graphviz-interactive-preview" // Graphviz DOT preview
+				"streetsidesoftware.code-spell-checker" // Catch spelling errors in docs
 			]
 		}
 	}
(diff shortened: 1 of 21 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "Development",
+	"name": "Development (GPU)",
```
guzman-raphael: These updates seem to be committed by mistake. Here you are updating the non-GPU DevContainer.

src/crypto.rs
```diff
@@ -0,0 +1,41 @@
...
+        reader.consume(buffer_len);
+    }
+
+    Ok(format!("{:x}", hasher.finalize()))
+}
+
+#[expect(dead_code, reason = "temporary for next pull request")]
+/// Function to hash data that is already in memory. This is much cleaner and less overhead compare
+/// to mapping data in memory into a ``BufReader``
+pub fn hash_bytes(data: impl AsRef<[u8]>) -> String {
```
guzman-raphael: I think it would be clearer if it were named:

```suggestion
pub fn hash_buffer(buffer: impl AsRef<[u8]>) -> String {
```

src/crypto.rs
```diff
@@ -0,0 +1,41 @@
...
+    let mut hasher = Sha256::new();
+    hasher.update(data);
+    format!("{:x}", hasher.finalize())
```
guzman-raphael: This can be simplified to this, no?

```suggestion
    format!("{:x}", Sha256::digest(buffer.as_ref()))
```

src/crypto.rs
```diff
@@ -0,0 +1,41 @@
+use sha2::{Digest as _, Sha256};
```
guzman-raphael, line 3: I'd suggest we add just one test in this PR. A valid one that comes to mind is to take a file in the tests (e.g. `tests/data/images/dog.jpeg`) and hash using both of these methods and verify that they arrive at the same hash.

It might make the `dead_code` exception not necessary. :shrug:

src/crypto.rs
```diff
@@ -0,0 +1,41 @@
...
+        };
+
+        if buffer_len == 0 {
+            break;
+        }
+        reader.consume(buffer_len);
+    }
+
+    Ok(format!("{:x}", hasher.finalize()))
+}
```
guzman-raphael: Is the `BuffReader` necessary? Why not just use `read_exact` directly?

```suggestion
pub fn hash_stream(stream: &mut impl Read) -> Result<String> {
    const SIZE: usize = 32;
    let mut hash = Sha256::new();
    let mut buffer: [u8; SIZE];
    while let Ok(_) = {
        buffer = [0; SIZE];
        stream.read_exact(&mut buffer)
    } {
        hash.update(buffer);
    }
    hash.update(buffer
        .into_iter()
        .filter(|x| *x > 0)
        .collect::<Vec<_>>());
    Ok(format!("{:x}", hash.finalize()))
}
```

guzman-raphael (changes requested):

src/crypto.rs
Synicix: I think that was a side affect of when I was using an entire different hashing system that didn't just do Sha256. Good catch

src/crypto.rs
Synicix: When I was naming it, I was thinking of it as raw bytes, since a lot of functions in string for example is like to_bytes.

src/crypto.rs
Synicix: Wasn't aware of this function, I can look into it

Synicix: Okay a quick look over it
```while let Ok(_) = {
        buffer = [0; SIZE];
        stream.read_exact(&mut buffer)
    } {
        hash.update(buffer);
    }
```

This doesn't exactly hash by chunks as I wanted it to. Basically the flow I was aiming for is read in 32 bytes -> update hash -> trash it and read the next 32 bytes. This allows it to read really big files without running into memory issues.

Thought the while let is new to me, and might be able to use that if it makes the code simpler

src/crypto.rs
Synicix: Fair enough, I can add that later today.

Synicix: Added tests

.devcontainer/devcontainer.json
Synicix: fixed

src/crypto.rs
guzman-raphael: Wait, isn't what I proposed doing the same thing? I allocated `buffer` and then the `read_exact` chunk overrides the previous value while updating the hash in-between i.e. discarding the prior value. Is it growing in memory in your test? That would be unexpected.

.devcontainer/devcontainer.json
guzman-raphael: Thanks but my comment applied to all the proposed changes to the entire file. Looks like you updated only this one line. :sweat_smile:

src/crypto.rs
guzman-raphael: I see. Technically though, this method can accept bytes directly but also anything that can be converted/accessed as bytes e.g. `&str`, `String`, and much more. In other words, pretty much anything in memory which is why `buffer` came to mind.

src/crypto.rs
Synicix: I see, I got confused with line 26-28. Didn't realized that syntax was even a thing. I reading it as the hash.update part doesn't run every loop

Synicix: Okay I I haven't got a chance to look at one issue into futher, but the hashing comes out differently possibbly due to this behavior
```
If this function encounters an “end of file” before completely filling the buffer, it returns an error of the kind [ErrorKind::UnexpectedEof](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.UnexpectedEof). The contents of buf are unspecified in this case.
```

https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact

I haven't gotton a chance to look into too deep, but that is so far my guess why it is resulting in a different hash.

guzman-raphael: I created a version in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0df55b82bc80afa04c79d4b03e9d557a) that compares them. All hashed to the name value.

Also, I'm going to conveniently do some hand-waving around Rust Playground's security. :upside_down_face:

Synicix: Hmm okay then it might be something on my side. Will check when I get back home

guzman-raphael (changes requested):
@Synicix Thanks for the update! I like that you found `read(..)` and how you used it to verify how many bytes were read. Feels like that was the missing piece.

The remaining concerns I have can be summarized as follows:
- `compute_checksum_for_path` (directory) doesn't consider the relative path. This is important since it can affect reproducibility when a directory is an input.
- `compute_checksum_for_path` feels busy with the various logic paths you've put into it. I'd suggest we separate it into explicit functions e.g. `hash_file` and `hash_dir`.
- Looking over `hash_stream` again, there is actually is a way to remove the need for `break`. 
- The `static` variable (seems like we only are using one) is only used by `hash_stream` and felt clunky to be separate i.e. violating functional programming design. The overhead is so small that I figured lets just put it in the function itself that way they stay together and it is easy for anyone else to extend/modify, future-proofing this method. For instance, I could see us later exposing the chunk size as an arg especially since it won't affect the hash value.
- The docs are a bit wordy and the tense isn't consistent. Could use a clean up pass.
- Seems like the `error::Kind` was reorded for some reason. I prefer the previous order since it was easy to spot the 'wrapped' errors and our custom ones.
- There are still unintentional commits to the non-GPU DevContainer env.

To help convey these ideas, I've also issued a [PR](https://github.com/Synicix/orcapod/pull/5) to your fork addressing these points.

guzman-raphael (approved):

PR #68, opened 2025-04-01
Fix diagram bug

guzman-raphael:
This PR addresses the following:
- Manually separate error logs to `stderr` on diagram docs utility due to bug
- Add diagram docs utility to GPU DevContainer
- Set Codespaces minimum resources config but let DevContainer consume all available by default
- Bump rust edition
- Fix format discrepancies due to edition update

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -20,8 +20,6 @@
 	},
 	"runArgs": [
 		"--name=${localWorkspaceFolderBasename}_devcontainer",
-		// "--cpus=2",
-		// "--memory=8gb",
 		"--privileged",
 		"--cgroupns=host"
 	],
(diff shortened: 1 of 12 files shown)


.devcontainer/devcontainer.json
```diff
@@ -20,8 +20,8 @@
 	},
 	"runArgs": [
 		"--name=${localWorkspaceFolderBasename}_devcontainer",
-		"--cpus=2",
-		"--memory=8gb",
+		// "--cpus=2",
```
Synicix: Should we go ahead and just remove it then?

Synicix (approved):
Aside from the question of just removing the comment. All changes look pretty straight forward and simple.

.devcontainer/devcontainer.json
guzman-raphael: Good point. I've moved it into a doc in our README to make it handy when needed.

Synicix (approved):

PR #70, opened 2025-04-03
`thiserror` -> `snafu`

guzman-raphael:
Depends on #69

In this PR, we:
1. Replace`thiserror` in favor of `snafu` for error handling
1. Add backtraces on all errors if available i.e. respect `RUST_BACKTRACE` environment variable
1. Inject additional contextual info on external errors that could use more help e.g. adding the path to `std::io::Error` when encountering `No such file or directory (os error 2)`
1. Remove usage of `OrcaError::from(Kind::..)` since unneeded. `?` automatically converts due to `impl From<Kind> for OrcaError {..}`.
1. Add error tests
1. Improve colors on warning messages
1. Add `cargo expand` macro expansion CLI (useful in debugging)
1. Fix diagram doc bug


Synicix (approved):
Reviewed by using it in failure mode PR

PR #71, opened 2025-04-05
Add UniFFI client with Python bindings

### LINKED ISSUE #75: Non-Rust user interface design (Python)

guzman-raphael:
Depends on #70, #74

UniFFI Note:
- An `Object` is when something is Rust-owned and passed by reference to the client. Methods are supported but indexing is only supported via methods e.g. getters.
- A `Record` is when you transfer ownership to the client and passed by value. Methods are not supported but the client provides native indexing support.
- An `Enum` is most similar to a `Record` above but for a Rust enum.

In this PR, we:
1. Add UniFFI dependency. It can generate the C FFI lib + the client-specific bindings e.g. Python. Included the `uniffi-bindgen` binary directly in our package since it simplifies the orcapod installation steps (users don't need to install it separately).
   1. Add `async-trait` dependency so `async fn` work with dyn traits.
   1. Add `getset` dependency which allows auto deriving getters for structs. Sourcing from my fork since needed to add a feature to propagate specified attributes (jbaublitz/getset#114).
   1. Add `derive_more` (`display`) to simplify how to modify `Display` implementations since there is good integration with UniFFI i.e. this is how you modify how a Rust-owned object is displayed on the client side.
1. Modify source.
   1. Expose the following in C lib using UniFFI.
       1. `Object`: `Pod`, `PodJob`, `OrcaError`, `LocalDockerOrchestrator`, `LocalFileStore`.
       1. `Record`: `PodResult`, `Annotation`, `GPURequirement`, `StreamInfo`, `OrcaPath`, `Blob`, `RunInfo`, `PodRun`, `ModelInfo`.
       1. `Enum`: `Model`, `GPUModel`, `Input`, `BlobKind`, `ImageKind`, `Status`, `ModelID`.
   1. Change any inheritance-like relationships to using `Arc` since UniFFI requires this for any `Object` usage e.g. `PodJob { pod: Pod, .. }` ->` PodJob { pod: Arc<Pod>, .. }`.
   1. Convert `delete_annotation` to non-generic since this is not allowed across CFFI boundary. Used a variation based on enum.
   1. Add `parse_debug_name`: a similar method to `get_type_name` but sourcing from the debug display. This is useful since it can be called with a struct or an enum.
   1. Change our opaque error from `OrcaError(Kind)` to `OrcaError { kind: Kind}` since UniFFI did not allow modifying display on tuple structs.
   1. Move `MODEL_NAMESPACE` constant out of trait since this is not supported in UniFFI.
1. Add Python utilities to IDE to help in building and testing the Python client.
1. Add `pyproject.toml` so others can easily install from source e.g. `pip install git+https://github.com/walkerlab/orcapod`.
1. Reorg non-Rust test data into a `extra/` subdirectory.
1. Add `tests/extra/python/smoke_test.py` to verify some basic features in Python client. Supports using debugger.
1. Add `docs/examples` and move diagram to `docs/images/`.
1. Clean up `.clippy.toml`.
   1. UniFFI generates some idents we don't have control over so set `min-ident-chars-threshold=1` (default).
   1. Update `allow-duplicate-crates` to current offenders


src/core/store/filestore.rs
```diff
@@ -51,12 +55,17 @@ impl LocalFileStore {
         PathBuf::from(format!("annotation/{name}-{version}.yaml"))
     }
     /// Build the storage path with the model directory (`hash`) and a file's relative path.
-    pub fn make_path<T>(&self, hash: &str, relpath: impl AsRef<Path>) -> PathBuf {
+    pub fn make_path<T: fmt::Debug>(
+        &self,
+        model: &T,
```
Synicix, line 60: What was the reason to change
```rust
pub fn make_path<T>(&self, hash: &str, relpath: impl AsRef<Path>) -> PathBuf
```
to include model: &T, when users of this function can still use 

```rust
::<T>make_path()

```

src/core/util.rs
```diff
@@ -14,6 +14,18 @@ pub fn get_type_name<T>() -> String {
         .unwrap()
 }
 
+#[expect(
+    clippy::unwrap_used,
+    reason = "Cannot return `None` since debug format always returns `String`."
+)]
+pub fn parse_debug_name<T: fmt::Debug>(instance: &T) -> String {
```
Synicix, line 21: What was the reason behind adding this function when it does similar to what get_type_name does?

src/uniffi/model.rs
```diff
@@ -8,13 +8,32 @@ use crate::{
...
+use derive_more::Display;
+use getset::CloneGetters;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, path::PathBuf};
+use std::{collections::HashMap, path::PathBuf, sync::Arc};
+use uniffi;
+
+/// Available models.
+#[derive(uniffi::Enum, Debug)]
+pub enum Model {
```
Synicix: Perhaps go with ModelType or Type?

tests/crypto.rs
```diff
@@ -8,7 +8,7 @@ use std::fs::read;
 
 #[test]
 fn consistent_hash() -> Result<()> {
-    let filepath = "./tests/data/images/subject.jpeg";
+    let filepath = "./tests/extra/data/images/subject.jpeg";
```
Synicix, line 11: What was the reason to add extra?

tests/error.rs
Synicix, line 1: Perhaps later we should match against the kind to see if it returns the correct error type?

tests/extra/python/smoke_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+def create_pod(data, _):
+    data["pod"] = Pod(
+        annotation=Annotation(
+            name="simple",
+            description="This is an example pod.",
+            version="1.0.0",
+        ),
+        image="alpine:3.14",
+        # command=r"sh -c 'sleep 5 && echo finished...'", # needs arg parsing before will work
```
Synicix: Stale comment code?

Synicix (commented):
Some discussion points for next meeting

guzman-raphael:
Patch coverage here is less than 100% but I think for this particular case it is reasonable so leaving it as-is. It is because of UniFFI.

Reasons:
1. UniFFI has not added [automatically_derived](https://github.com/rust-lang/rust/pull/120185) to all generated `impl` cases which means that the generated code isn't omitted from coverage. Fixing this would require us to enhance UniFFI and use our own forked version of UniFFI until they accept the improvement.
1. Omitting coverage from generated functions (for us these are from UniFFI) is only possible by moving them to a specific file that can be omitted on Rust stable which would be too disruptive simply just for coverage. On Rust nightly, this [feature](https://github.com/rust-lang/rust/issues/134749) seems relevant but even then, it would require us to enhance UniFFI and use our own forked version of UniFFI until they accept the improvement.

The benefit doesn't seem worth it right now so I recommend we defer.

Synicix (approved):
Reviewed with zoom meeting

PR #80, opened 2025-04-23
Improve diagram generation

guzman-raphael:
Since there have been recent issues with the diagram generation, thought it would be best to properly include it into our tests to prevent finding out about an issue **after** merge.

To that end, in this PR we:
- Make the diagram build step a reusable action so it can be used as a test and when publishing as a doc
- Simplify the diagram output to be less confusing
- Upload to the GHA run the generated diagram image during the diagram test as an artifact
- Rename the base tests from `Tests/tests` -> `Tests/syntax-style-format-and-integration`
- Rename diagram image to `orcapod_diagram.svg` so downloading the artifact is less confusing


Synicix (approved):
Looks good

PR #83, opened 2025-05-19
Rough Draft Implementation of Pipeline, PipelineJob, PipelineRun, and PipelineRunner

Synicix:
Added:
- Pipeline struct (Still missing hashing and to yaml)
- PipelienJob (Still missing hashing, plus additional fields that we might want to add
- PipelineRun
- DockerPipelineRunner: An async task based runner of pipelines

TODO:
General: 
- Add to yaml and hashing to the above
- Create the storage function needed to store results
- Formalize the yaml to match with @guzman-raphael design
- Tag is completely missing

Pipeline:
- Add functionailty to actually create the pod job and send it to the orchestrator (Needs discussion to figure out, the details of what to have the pod_job be)
- Need to add a more proper pipeline unlike the toy example included in the draft at the moment
- How to deal with multiple streams

DockerPipelineRunner:
- Add functionailty to save pod results
- Need to add test case to test mutiple parents
- Need to add test case to test mutiple children
- Discuss what functions we want to have in detail, do we just copy docker orchestrator somewhat?

Known Bugs:
- Rust 1.87.0? update seem to screw with clippy and rust analyser, not sure how to fix it but current it is throwing incorrect warnings for test and one part of the code.

Other pending question:
- Should we add an option for the pipeline executor accept pod jobs and run and wait on them to save the pod result like a server?


Synicix:
Closing since this is stale and need to be split up

PR #85, opened 2025-05-27
Prototype: Add orchestrator agent

guzman-raphael:
This PR is not meant to be merged but a space to prototype several design concepts. Proper features will be polished out and introduced in a series of separate PRs.

todos that could be deferred to separate issues:
- Optimize Tokio runtime to be created once globally and reused everywhere
- Make the agent generic based on `Orchestrator` and `Store` traits to keep implementation clean e.g. `Arc<dyn Ochestrator>` seems to be supported by UniFFI but needs investigation to determine why it isn't working properly.
- Remove `unwrap()` usage in `deserialize_*` methods.
- Simplify `PodJob` state in orchestrator tags since JSON now includes expanded refs i.e. saving `Pod` separately is unneeded.
- Cleanup host name references e.g. agent's should refer to their host names but should clients refer to a remote host name or their own unique client name?
- Figure out how to properly do error handling to achieve `.lock()?` for `Mutex` with `snafu`.
- Investigate RwLock vs Mutex...


guzman-raphael:
Closing since this was mainly to protoype some ideas before formally introducing the new features. May still be useful to refer back to in future for the features that didn't make the cut in our current milestone.

PR #88, opened 2025-06-18
Rename parameters to match new spec

### LINKED ISSUE #87: Rename pod and pod job struct to match new standard
Image

Image

Summary:

Pod Job:
Input Stream -> Input Packet

Pod
Input -> PathSet
StreamInfo -> PathInfo
input_stream -> input_spec
output_stream -> output_spec

Other Stuff:
OrcaPath -> URI

Synicix:
Just a simple rename, nothing complicated.
- #87


src/uniffi/model.rs
```diff
@@ -284,7 +284,7 @@ pub struct StreamInfo {
 /// Input options.
```
guzman-raphael: This doc should be updated.

tests/fixture/mod.rs
```diff
@@ -80,7 +80,7 @@ pub fn pod_job_style(namespace_lookup: &HashMap<String, PathBuf, RandomState>) -
         HashMap::from([
             (
                 "extra-style".to_owned(),
-                Input::Unary(Blob {
+                PathSet::Unary(Blob {
                     kind: BlobKind::File,
                     location: OrcaPath {
```
guzman-raphael: Could you also rename `OrcaPath` -> `URI` everywhere?

guzman-raphael (changes requested):
Thanks @Synicix! :clap: 

Left some feedback. Aside from that, just need to update `tests/extra/python/smoke_test.py`.

tests/fixture/mod.rs
Synicix: Opps, missed that from the notes. Thanks

Synicix:
@guzman-raphael  Applied fixes requested

guzman-raphael (changes requested):
@Synicix Thanks for the update.

Looks like the CI is broken due to the `clippy::multiple_crate_versions`  lint again... 

Could you either:
- add
  ```toml
  multiple_crate_versions = { level = "allow", priority = 127 } # allow since list of exceptions changes frequently from external
  ```
  to `Cargo.toml`
- depend on #90

Since it happens so frequently, I think it is time to finally ignore that lint.

Synicix:
@guzman-raphael  I can add those two. Should probably review and push your thing in first then.

Synicix:
@guzman-raphael Update the python smoke test to work with the new names, and merge the changes from your other pull request in.

guzman-raphael (approved):

PR #90, opened 2025-06-19
Add multiple serialization schemes for models

guzman-raphael:
- Add multiple serialization schemes for models e.g. YAML => storage optimized, JSON => memory/comm optimized
- Use global tokio runtime when calling `async` from `sync`
- Fix bug when calling async functions in Python using `asyncio`
- Improve `smoke_test.py` exit code checks
- Bump Python p39 -> py310 since debugger requires it
- Add Python integration tests to CI
- Bump Rust
- Exclude clippy checks for `multiple_crate_versions` since typically caused by external dependencies that are out of our control

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,19 +1,3 @@
 excessive-nesting-threshold = 4
 too-many-arguments-threshold = 10
 allowed-idents-below-min-chars = ["..", "k", "v", "f", "re", "id", "Ok", "'_"]
-allowed-duplicate-crates = [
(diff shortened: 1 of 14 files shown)


Synicix:
Adding to the list for completeness.

- Change orchestrator to encode the pod job into the labels of containers instead of annotation + hash + pod

Synicix (approved):
Looks good. I am going ahead and merge this @guzman-raphael @eywalker.

PR #91, opened 2025-06-20
Minimal orchestrator agent w/o reservations

guzman-raphael:
Depends on #90 

- Add an orchestrator `Agent` to manage requests for `PodJob` execution. This is meant to be "installed" by an administrator who has access to the orchestrator control plane. Currently, does not coordinate with other installed (i.e. running) agents (will be supported once reservation feature is in).
- Add a `AgentClient` that allows connecting to any `Agent` orchestration node to allow submitting requests remotely. [Zenoh](https://zenoh.io/) will reroute communication based on optimal path to distributed network.

Known places for improvement. Propose we file the following into separate issues:
- `agent.start(..)` currently is hard coded to `Arc<LocalDockerOrchestrator>` as opposed to intended `Arc<dyn Orchestrator>`. This is due to a UniFFI issue. We can probably live with this for a bit until properly fixing it. A likely solution could be in exporting `Orchestrator` using `#[uniffi::export(with_foreign)` but it has consequences i.e. more time is needed to properly address compiling errors.


Cargo.toml
```diff
@@ -37,6 +37,8 @@ chrono = "0.4.39"
 colored = "2.1.0"
 # derive utilities for new types
 derive_more = { version = "2.0.1", features = ["display"] }
+# allow sync functions to be called from async using `block_on`
+futures-executor = "0.3.31"
```
Synicix, line 41: Isn't this already handle by the tokio or sync lib?

src/core/orchestrator/agent.rs
```diff
@@ -0,0 +1,196 @@
...
+)]
+pub async fn start_service<
+    EventClassifierF,
+    RequestF,
+    RequestI,
+    RequestR,
+    ResponseF,
+    ResponseI,
+    ResponseR,
+>(
```
Synicix: Without docs and the variable name F I R, it is difficult to figure out what is going on. Perhaps better docs and names would help.

src/core/orchestrator/agent.rs
```diff
@@ -0,0 +1,196 @@
...
+                    serde_json::from_slice::<RequestI>(&sample.payload().to_bytes()),
+                    RE_PODJOB_ACTION.captures(sample.key_expr().as_str()),
+                ) {
+                    let inner_response_tx = response_tx.clone();
+                    let event_metadata = EventMetadata {
+                        group: metadata["group"].to_string(),
+                        host: metadata["host"].to_string(),
+                        subgroup: metadata["pod_job_hash"].to_string(),
+                    };
+                    let _event_payload = event_classifier(&input);
```
Synicix, line 166: Not sure why this is called but the result is discarded

src/uniffi/orchestrator/mod.rs
```diff
@@ -65,7 +65,7 @@ pub struct PodRun {
 /// API for standard behavior of any container orchestration engine supported.
 #[uniffi::export]
 #[async_trait::async_trait]
-pub trait Orchestrator: Send + Sync {
+pub trait Orchestrator: Send + Sync + fmt::Debug {
```
Synicix, line 68: Is fmt needed? Perhaps Debug is just good enough?

Synicix (approved):
Touch bases

Synicix (commented):
For discussion

Cargo.toml
Synicix: Leave it for now since it a work around for nested runtime.

Synicix:
From Discussion on zoom:

- Answered all related questions about designed, and all decisions looks good.
- @guzman-raphael Will look into possible replacement of future_ext lib with tokio::runtime::Handle

PR #94, opened 2025-06-26
`Agent`: Add configurable storage of `PodResult`

### LINKED ISSUE #81: Calling get_result_blocking immediately after starting container can result in container not found error
There seems to be an issue with calling get_result_blocking() right after start_blocking() as it errors out with container not found.

The underlying function called is wait_container_reponse unable to find the container despite it being started. I am suspecting because it takes time for the container to start for us to be able to wait on it. Perhaps we need to rethink this approach or use a different wait to wait for completion.

The current behavior is extremely inconsistent, it seems to fail when running cargo test, but individually test run doesn't seem to cause the problem. Also repeated runs after the first run seems to solve the issue, perhaps due to caching.

I will follow up on this issue and resolve it another time.

Pulling from one of the failed actions logs:
```
Warning: Docker container wait error
Warning: Docker container wait error
test fail_at_start ... ok
test command_parse ... FAILED
Error: Docker responded with status code 404: No such container: unhealthy-pies
stack backtrace:
```

Seems that our orcapod code spits out ``` Warning: Docker container wait error``` right before the main error ```Error: Docker responded with status code 404: No such container: unhealthy-pies```

Full failed action logs: https://github.com/walkerlab/orcapod/actions/runs/14921920603/job/41918758357

guzman-raphael:
Depends on #91 

- Add `available_store` option to `agent.start(..)` which will save pod results as they become available.
- Use a dedicated `Response` enum as opposed to casting async `Ok(())` responses to string over FFI boundary.
- Fix bug where `orch.get_result(...)` sometimes returned a `PodRun` with `Status::Running`.
- Fix bug where `orch.get_result(..)` would throw an error if a `PodRun` was unsuccessful instead of producing a `PodResult`  i.e. storing failed pod results can improve the user debugging experience.
- Fix bug related to some collisions between agent and orchestrator tests.
- Include `ContainerStateStatusEnum::REMOVING` when determining how to cast to `uniffi::orchestrator::Status`.

Known places for improvement. Propose we file the following into separate issues:
- Similar to #91, `agent.start(..)` adds another hard coded `Arc<LocalFileStore>` as opposed to the intended `Arc<dyn Store>`. That PR has a more detailed description and the reasoning here is the same. Once we identify a good solution, both can be fixed together.

### DIFF
--- a/tests/orchestrator.rs
+++ b/tests/orchestrator.rs
@@ -129,6 +129,7 @@ fn remote_container_image_basic() -> Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[expect(clippy::use_debug, reason = "Useful in debugging from CI.")]
 async fn verify_pod_result_not_running() -> Result<()> {
     let results = join_all(
         pod_jobs_stresser(
--- a/tests/orchestrator.rs
+++ b/tests/orchestrator.rs
@@ -36,6 +36,7 @@ where
         orchestrator
             .list_blocking()?
             .iter()
+            .filter(|container| container.pod_job == pod_run.pod_job)
             .map(|run| Ok(orchestrator.get_info_blocking(run)?.command))
             .collect::<Result<Vec<_>>>()?,
         vec![expected_command.clone()],
@@ -52,6 +53,7 @@ where
         orchestrator
             .list_blocking()?
             .iter()
+            .filter(|container| container.pod_job == pod_run.pod_job)
             .map(|run| Ok(orchestrator.get_info_blocking(run)?.command))
             .collect::<Result<Vec<_>>>()?,
         vec![expected_command],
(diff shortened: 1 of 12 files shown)


tests/orchestrator.rs
```diff
@@ -63,7 +69,10 @@ where
     // test delete
     orchestrator.delete_blocking(&pod_run)?;
     assert!(
-        orchestrator.list_blocking()?.is_empty(),
+        !orchestrator
+            .list_blocking()?
+            .iter()
+            .any(|container| container.pod_job == pod_run.pod_job),
```
Synicix, line 75: Perhaps we should use filter by pod_run since that would be unique than pod_job. We did discuss this a while back and I believe one of the standing PR does resolve this, so up to you if we should merge it as is.

Synicix (approved):
Pretty straight forward. There is just one comment, and it would be ready for merge.

PR #97, opened 2025-06-30
Add additional error handling for crypto.rs and accompany tests

Synicix:
- Rename InvalidFilepath to InvalidFileorDirPath error
- Added error handling that returns the dir when the path is invalid in hash_dir
- Added a test to cover the case of nested directory hashing

Fixes PLT-234


src/core/store/filestore.rs
```diff
@@ -194,7 +194,7 @@ impl LocalFileStore {
                 Ok((
                     serde_yaml::from_str(
                         &fs::read_to_string(path.clone())
-                            .context(selector::InvalidFilepath { path })?,
+                            .context(selector::InvalidFileOrDirPath { path })?,
```
eywalker: Why not just call this `InvalidPath`?

Synicix: True that is a better name, can issue the change in a bit

eywalker: Do you want me to merge as is and you change it in other PR or make changes in this PR?

Synicix: I an update it then you can merge it

Synicix: Updated

eywalker (approved):

eywalker (approved):

PR #99, opened 2025-07-01
MVP Pipeline Implementation with Utils update

Synicix:
Summary:
In short, I implemented from_dot to fit with the design of graph + kernel_lut where kernel_lut contains only the unique kernels to prevent duplicates. This will provided me with a minimum Pipeline for me to resume work on the pipeline runner. Also, it removes the need for pipeline_builder to be implemented in order to create test cases.

Added: 
- MVP pipeline
- from_dot function that requires the graph and kernel_to_node_name HashMap  (Reason behind this is I still want to keep kernel_lut to only have 1 copy of each kernel)
- find_missing_keys, for PipelineJob input checking
- Uniffi bindings

Updated:
- utils functions to more generic implementation

Deferred: 
- to_dot (Was spending too much time trying to deal with error handling based on @guzman-raphael reference code, and it is not needed for runner)
- to_svg (Not needed for runner so deferring it)

For reference where these changes come from:

image


Synicix:
@guzman-raphael  The code base is ready for a rough review, it is still missing tests and the list of todos above, but I plan to get it done by Wednesday night hopefully.

Cargo.toml
```diff
@@ -49,6 +49,7 @@ heck = "0.5.0"
 indexmap = { version = "2.9.0", features = ["serde"] }
 # random name generator
 names = "0.14.0"
+petgraph = { version = "0.8.2", features = ["serde-1"] }
```
guzman-raphael: Before introducing `petgraph` serialization (assuming for storage), we should ensure it is deterministic.

e.g.:

```mermaid
graph TD;
    A-->B;
    A-->C;
```

and

```mermaid
graph TD;
    A-->C;
    A-->B;
```

should serialize the same in YAML. If not, we should defer until we can guarantee this.

src/core/mod.rs
```diff
@@ -4,5 +4,7 @@ pub(crate) mod error;
 /// Components of the data model.
 pub mod model;
 pub(crate) mod orchestrator;
+/// Components relating to pipelines
+pub mod pipeline;
```
guzman-raphael: We should not public expose here. If meant to be public, it should be moved into within `orcapod::uniffi::*` with appropriate FFI exposure.

src/uniffi/model.rs
```diff
@@ -304,6 +304,20 @@ pub struct Blob {
     /// BLOB contents checksum.
     pub checksum: String,
 }
+impl Blob {
+    /// Construct a new BLOB instance.
+    ///
+    /// # Errors
+    ///
+    /// Will return `Err` if there is an issue initializing a `Blob` instance.
+    pub const fn new(kind: BlobKind, location: URI) -> Self {
```
guzman-raphael: Why was this needed? Why not just: `Blob {kind, location, ..Blob::default()}`.

Synicix: Actually, very fair point, let me fix this.

src/core/mod.rs
Synicix: Add TODO move it to uniffi side of the library, and add to derives to function and struct that we want to export.

Cargo.toml
Synicix: File an issue to properly fix this issue. At the moment it is none deterministic. 

- TODO: Disable hash for now.

src/core/mod.rs
Synicix: Added

guzman-raphael (changes requested):
@Synicix I have completed my review of your PR and summarized my feedback into this [PR](https://github.com/Synicix/orcapod/pull/7) to your fork. Let me know if you have any questions when you go over it.

I'd recommend merging it through GH to prevent merge conflicts between us and so my contributions show up in the commit history. If you'd like to make some changes yourself, it is probably best to do it after merging my PR to your fork.

guzman-raphael (approved):

PR #102, opened 2025-07-24
Fix OOM issue with stress test

Synicix:
- Short fix due to the problem I experienced with swapless machines.


guzman-raphael:
Closing since #99 included this [fix](https://github.com/walkerlab/orcapod/pull/99/files#diff-f45ad32b9e90e5cc5baa085cb104208783134c08108158fb691832e9554586acR163-R182) and it was recently merged.

PR #105, opened 2025-07-30
Make `Pod`'s command more flexible

### LINKED ISSUE #95: Change command from string to a vec of string

guzman-raphael:
## Feature
- Fix #95

## Housekeeping
- Rename `submit_pod_jobs` -> `start_pod_jobs`.
- Simplify response task for pod job requests.

## Docs
- Add enum hint doc for uniffi.

## Tests
- Improve debug experience when using agent client watch.

### DIFF
--- a/src/core/orchestrator/docker.rs
+++ b/src/core/orchestrator/docker.rs
@@ -130,12 +130,6 @@ impl LocalDockerOrchestrator {
             ),
             ("org.orcapod.pod_job.hash".to_owned(), pod_job.hash.clone()),
         ]);
-        let command = pod_job
-            .pod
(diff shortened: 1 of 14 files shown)


Synicix (approved):
Simple update looks good.

PR #107, opened 2025-08-02
Enable using Git Worktrees

guzman-raphael:
Recently I've learned about [Git Worktrees](https://git-scm.com/docs/git-worktree). It is a feature somewhere between [Git Stash](https://git-scm.com/docs/git-stash) and completely copying/cloning a repo again.

It is great for context switching and superior to stash because it:
- Creates a separate directory on your machine.
- Synchronizes Git between them without duplicating Git history i.e. ensures it is up to date and prevents collisions.

Because of this setup, you can launch separate DevContainers in parallel. This is useful because:
- It is less disruptive if you are deep into some uncommitted changes.
- You have access to rust analyzer/clippy e.g. when reviewing other PRs.
- It makes it easier to work on refactor/migrations between branches/stashes with separate tooling e.g. tests, runable notebooks, etc.
- It allows vibe coding iteration to continue in the background.

However, some changes were necessary to make our setup compatible with this.

## Pre-requisites

- Update your `git` to at least version `2.48` since that is when the `--relative-paths` option was added to `git worktree`.
- Create worktrees on your machine using the following command.
  ```bash
  git worktree add /path/to/new/dir-name branch-name --relative-paths # create
  git worktree remove dir-name # delete
  ```

## Changes

- Convert default DevContainer to compose since it allows setting DevContainer config dynamically.
- Switch DevContainer image since installing latest git is easiest on ubuntu.
- Add DevContainer prelaunch script that determines appropriate mount points.


Synicix (approved):
Discussed over zoom, and everything looks good

PR #110, opened 2025-08-10
Make zenoh key expression structure easier to maintain

guzman-raphael:
## Features
- Localize zenoh key expression structure to a single builder for both pub/sub (see `make_key_expr`). Keys are alphabetically sorted in key expression.
- Relax structure to be mostly self-describing. This means that the builder structure is minimal and the caller is free to add as much context as necessary via `metadata`.
- When in sub mode, key are delimited by wildcards to allow room to grow. The result is a key expression parsing logic that will be backward compatible with future releases provided we primarily add new keys in the future. Renames will present backward incompatibility but that is somewhat expected. Therefore to avoid this, we should try to pick good key names that we don't expect to rename in the future.
- Eliminate the need to maintain an agent regex and the capture struct i.e. `EventMetadata`.

## Tests
- Add a `default` and a `test` feature. It is only allowed to enable one of them at a time.
  - Default (which is the default feature) will not expose any part of the private API in `core`.
  - Test will expose some portion of `core`. This is useful when integration tests need access to `core`.

## Docs
- Add note about using `git worktree`

## Housekeeping
- Simplify errors related to `None` values to a single `MissingInfo` error. The `details` field can be used to elaborate on what is missing and provide a hint to API user.
- Clean up some of the `while` loops that silenced certain error states.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Install code coverage
         uses: taiki-e/install-action@cargo-llvm-cov
       - name: Run syntax and style tests
-        run: cargo clippy --all-targets -- -D warnings
+        run: cargo clippy --no-default-features --features=test --all-targets -- -D warnings
       - name: Run format test
         run: cargo fmt --check
       - name: Run integration tests w/ coverage report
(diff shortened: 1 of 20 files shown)


Synicix (approved):
Merge to have a base to work off of

PR #114, opened 2025-08-29
Splitting test into unit and integration tests to match with Rust testing standards

### LINKED ISSUE #113: Split test into integration test (test folder) vs unit testing
Reverting the test feature of crate to be more align with rust traditional testing standards.

Synicix:
Depends on #111 

After taking a look at using cargo features to allow testing. I believe this strays too far from the standard way Rust expects us to test things. In short I am splitting them based on the purpose of the test.

- Test folder (Integration test, testing usage of orcapod from the user side)
- Integrated test mods in the actual code files (For unit and internal test): Note the compiler does not compile the unit test unless you run cargo test as documented here: https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest

Resolves PIPE-118


eywalker (approved):

eywalker:
Have reviewed through in-person discussion. This, however, warrants a more thorough future visit upon completely cleaning up the series of PRs.

PR #115, opened 2025-09-02
Update pod hashing to remove non-replication concerns from hash

### LINKED ISSUE #57: Source commit url should be moved out of Pod spec
We decided on perhaps moving it to a specific PodAnnotation that is seperate from Annotation. Should make it optional

Synicix:
- Remove source commit for now (Perhaps we should discuss to add a additional_label member where it is HashMap<String, String> and the user can put whatever they want in there. This would be far more flexible and allow for custom labels
- Remove recommended_cpu and recommended_memory and combined it into RecommendSpecs struct that no longer affects the hash of the pod
- Update GPU requirement specs as that is very code dependent and is a all or nothing when it come to reproducibility
- Clean up a few things I caught to make the code base more robust

Depends on #114 

Resolves PIPE-119


eywalker:
Indeed, we should turn your points into issues to be tracked. Namely:
1. add `additional_label` so that users can add arbitrary information. Among these are Git hashes. Generally we should discuss the topic of capturing the concept of execution environment more broadly.
2. consider placing annotation about expected reproducibiility (e.g. GPU-based code can have not perfectly reproducible execution).

src/core/operator.rs
eywalker, line 1: We should most definitely turn this into a module so we can have smaller src code dedicated to each operator, especially as we add operators in the future.

eywalker (approved):

PR #116, opened 2025-10-02
Pipeline Validation

Synicix:
- Added validation features for pipeline creation

Depends on #115 

Resolves PIPE-120


src/core/model/pipeline.rs
```diff
@@ -0,0 +1,313 @@
...
+                })
+            });
+
+            input_keys.collect()
+        } else {
+            parent_indices
+                .into_iter()
+                .map(|parent_idx| {
+                    // Check if hash has been computed for this node, if not trigger computation
+                    if graph[parent_idx].hash.is_empty() {
```
eywalker: Would be good to move the hash check to the node itself instead of performing the hash property check on the parent node.

eywalker (changes requested):

src/core/model/pipeline.rs
Synicix: Fixed

eywalker (approved):

PR #122, opened 2025-10-23
Merge CPU and Dev container to share code and configuration files

Synicix:
Merge the two CPU and GPU dev containers to share as much common resources as possible.

Fixes PLT-298


.devcontainer/gpu/devcontainer.json
eywalker, line 1: Actually why do we need GPU-enabled container for development for Orcapod? Do we have something we specifically test GPU features?

Synicix: Not yet, but soon. I could also add the GPU feature and a the test for it.

eywalker: If we don't currently need GPU-based containers, I'd be in favor of removing it? How did you plan on using the GPU image?

eywalker: Anyway I'll go ahead and merge but we will follow up in a separate issue to discuss and simplify the devcontainer.

eywalker (approved):

.devcontainer/gpu/devcontainer.json
Synicix: I was planning to add a test to check if the Pod requires a gpu that it launches with one. The issue I ran into is that github action doesn't have GPU's by default so those GPU tests would fail. Since a lot of people will use this for machine learning, testing if orca pod handle GPU Pods correctly is needed.

Synicix: https://docs.github.com/en/actions/concepts/runners/larger-runners

I did find this a while ago but I am not sure if it would cost us more to get access to a GPU Github Action Runner.

The other option is to write a test as a switching code, where it will just return success if there is no GPU, and if there is a GPU available, then actually test it.

